PR #6, opened 2018-05-06
Update README.md

adamant-al:
Few changes


RealBonus:
Committed into develop with fixed markup.

PR #45, opened 2018-08-08
Feature/ipad ui

boyarkin-anton:
Use UISplitViewController for iPad UI


boyarkin-anton:
Merged into feature/Improvements branch.

PR #46, opened 2018-08-30
fix & update eth transaction sending

boyarkin-anton:
Send ETH transaction hash in chat before sending to Ethereum node
Update web3swift to latest version


RealBonus:
Merged in different branch.

PR #76, opened 2022-04-07
Processing qr codes and universal links

StanislavDevIOS:


Adamant/Utilities/AdamantUtilities+extended.swift
```diff
@@ -20,6 +20,10 @@ extension AdamantUtilities {
         return ""
     }()
     
+    static var applicationReleaseVersion: String = {
```
adamant-al, line 23: Remove this and use version from app build

adamant-al (approved):

adamant-al (approved):

PR #204, opened 2023-03-22
[trello.com/c/wUTl8kfK] Fixed ADM address markdown

StanislavDevIOS:


Adamant/Services/AdamantDialogService.swift
```diff
@@ -235,35 +235,56 @@ extension AdamantDialogService {
...
             preferredStyle: .actionSheet,
             source: from
         )
         
+        let source: ViewSource?
+        if let from = from {
+            source = .view(from)
+        } else {
+            source = nil
+        }
```
Andrey1029, line 259: `let source: ViewSource? = from.map { .view($)) }`

PR #228, opened 2023-04-26
[trello.com/c/yt09l4Xt] Start new chat from search

StanislavDevIOS:


Adamant/Stories/ChatsList/SearchResultsViewController.swift
```diff
@@ -11,35 +11,58 @@ import Swinject
...
+        router: Router,
+        avatarService: AvatarService,
+        addressBookService: AddressBookService,
+        accountsProvider: AccountsProvider
+    ) {
+        self.router = router
+        self.avatarService = avatarService
+        self.addressBookService = addressBookService
+        self.accountsProvider = accountsProvider
+        super.init(nibName: "SearchResultsViewController", bundle: nil)
```
Andrey1029, line 57: Its better to use `String(describing: Self.self)` to avoid literals

Adamant/Stories/ChatsList/SearchResultsViewController.swift
```diff
@@ -168,6 +201,34 @@ class SearchResultsViewController: UITableViewController {
...
+        guard let partner = partner else { return }
+        
+        if let avatarName = partner.avatar, let avatar = UIImage.init(named: avatarName) {
+            cell.avatarImage = avatar
+            cell.avatarImageView.tintColor = UIColor.adamant.primary
+        } else {
+            if let address = partner.publicKey {
+                let image = self.avatarService.avatar(for: address, size: 200)
+                
+                DispatchQueue.onMainAsync {
```
Andrey1029, line 214: Why is it here? This method is always on the main thread

Andrey1029 (approved):

PR #265, opened 2023-06-08
[trello.com/c/D11z4W4Y] Fixed open push notification

StanislavDevIOS:
- Implemented the feature to automatically scroll to the message or to the bottom of the chat if there are no messages when opening a push notification.
- Resolved a bug specific to iPad (and potentially macOS) where, upon the app becoming active, new messages would start loading even if the user hadn't initialized the loading.

### DIFF
--- a/Adamant/Stories/Chat/ViewModel/Models/ChatStartPosition.swift
+++ b/Adamant/Stories/Chat/ViewModel/Models/ChatStartPosition.swift
@@ -10,6 +10,5 @@ import CoreGraphics
 
 enum ChatStartPosition {
     case offset(CGFloat)
-    case messageId(String)
-    case messageIdOrBottom(String)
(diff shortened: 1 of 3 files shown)


Adamant/Stories/Chat/ViewModel/Models/ChatStartPosition.swift
```diff
@@ -11,4 +11,5 @@ import CoreGraphics
 enum ChatStartPosition {
     case offset(CGFloat)
     case messageId(String)
+    case messageIdOrBottom(String)
```
Andrey1029, line 14: When is it used?

Andrey1029:
```
case offset(CGFloat)
case messageId(String, toBottomIfNotFound: Bool)
```

Andrey1029 (approved):

PR #313, opened 2023-08-18
[trello.com/c/zlEHFxGJ] Fixed wrong node status

StanislavDevIOS:
Issue: The non-responsive node was taking a long time to respond (the default timeout is set quite high). On the screen with the list of nodes, there is a refresh timer that runs every 10 seconds. When all nodes are being refreshed, old requests are being cancelled and new ones are being created. However, since the previous request couldn't wait for a response and was cancelled, it led to an "explicitly cancelled" error. Such requests were being ignored (the status wasn't being assigned as offline).

Solution: I have adjusted the default timeout to 15 seconds. I've also modified the update logic: instead of cancelling all requests and creating new ones, I now only recreate the requests that have already completed. Requests that are still waiting for a response are left intact.

### DIFF
--- a/Adamant/Services/AdamantHealthCheckService.swift
+++ b/Adamant/Services/AdamantHealthCheckService.swift
@@ -18,7 +18,7 @@ final class AdamantHealthCheckService: HealthCheckService {
     // MARK: - Properties
     
     private var _nodes = [Node]()
-    private var currentRequests = Set<DataRequest>()
+    private var currentRequests = Set<URL>()
     private let semaphore = DispatchSemaphore(value: 1)
     private let notifyingQueue = DispatchQueue(label: "com.adamant.health-check-notification")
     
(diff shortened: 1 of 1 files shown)


Andrey1029 (approved):

Adamant/Services/AdamantHealthCheckService.swift
```diff
@@ -115,8 +124,10 @@ final class AdamantHealthCheckService: HealthCheckService {
     }
     
     private func resetRequests() {
-        currentRequests.forEach { $0.cancel() }
-        currentRequests = []
+        currentRequests.filter { $0.isFinished }.forEach {
+            $0.cancel()
+            currentRequests.remove($0)
+        }
```
Andrey1029, line 130: Тут нарушение логики. Бессмысленно отменять завершенный запрос.

Предлагаю вообще отказаться от хранения запросов и просто проверять, по каким нодам ответы пришли, а по каким нет.

Andrey1029 (changes requested):
.

Andrey1029 (approved):

PR #379, opened 2023-11-21
[trello.com/c/LRsSGO6R] Show height node in list

StanislavDevIOS:


just-software-dev (changes requested):
Let's cancel PR because these changes are already in another PR

PR #397, opened 2023-11-29
[trello.com/c/H3H8SXeM] Eth crash fix

just-software-dev:


Adamant/Modules/Wallets/Ethereum/EthWalletService+RichMessageProviderWithStatusCheck.swift
```diff
@@ -71,7 +71,7 @@ private extension EthWalletService {
     func getTransactionInfo(hash: String, web3: Web3) async throws -> EthTransactionInfo {
         try await withThrowingTaskGroup(
             of: EthTransactionInfoElement.self,
-            returning: EthTransactionInfo.self
```
just-software-dev, line 74: Yes, that's too much, but let's keep it in the 3.3.0 release to guarantee that a crash will not appear hear. Most ETH crashes appeared in this place

PR #421, opened 2024-01-17
[trello.com/c/rbrULUxL] Create own realisation for date

IanaaDvlp:


just-software-dev (changes requested):
https://developer.apple.com/documentation/xcode/localizing-strings-that-contain-plurals

just-software-dev (approved):

PR #458, opened 2024-02-29
[trello.com/c/QCpCeY3L] Feat: Decline second crypto transfer if we have a pending Tx

IanaaDvlp:
Feat: Decline second crypto transfer if we have a pending Tx (where no nonce: BTC, DASH, DOGE)


Adamant/Modules/Wallets/Dash/DashWalletService.swift
```diff
@@ -453,10 +453,23 @@ extension DashWalletService {
...
         }
         
+        coinStorage.append(trs)
+        
+        return trs.count
+    }
+    
+    func getTransactionsHistory(offset: Int, limit: Int) async throws -> [TransactionDetails] {
+        guard let address = wallet?.address else {
+            return []
```
just-software-dev, line 470: If there are no address — throw an error

Adamant/Modules/Wallets/Doge/DogeTransferViewController.swift
```diff
@@ -46,6 +46,18 @@ final class DogeTransferViewController: TransferViewControllerBase {
...
+                if await !readyToSendFunds() {
+                    dialogService.dismissProgress()
+                    dialogService.showAlert(
+                        title: nil,
+                        message: String.adamant.transfer.pendingTxError(coin: service.tokenSymbol),
+                        style: AdamantAlertStyle.alert,
+                        actions: nil,
+                        from: nil
+                    )
+                    return
```
just-software-dev, line 58: I see this code in every coin VC — can we move it to the base VC?

PR #462, opened 2024-03-05
[trello.com/c/BT9aPU32] Open the keyboard if inputText is not empty

IanaaDvlp:


Adamant/Modules/Chat/ViewModel/ChatViewModel.swift
```diff
@@ -188,6 +189,11 @@ final class ChatViewModel: NSObject {
         }
     }
     
+    func presentKeyboardIfNeeded() {
```
IanaaDvlp, line 192: presentKeyboardOnStartIfNeeded

PR #498, opened 2024-07-13
[trello.com/c/GJ5CEZSz] fix: dates not update

IanaaDvlp:


Adamant/Modules/Chat/View/ChatViewController.swift
```diff
@@ -333,6 +333,7 @@ private extension ChatViewController {
             .sink { [weak self] in
                 if $0 {
                     self?.updatingIndicatorView.startAnimate()
+                    self?.viewModel.refreshDateHeadersIfNeeded()
```
just-software-dev, line 336: Move it here (ChatViewModel):
```
await chatsProvider.stateObserver
                .receive(on: DispatchQueue.main)
                .sink { [weak self] state in
                    self?.isHeaderLoading = state == .updating ? true : false
                }
                .store(in: &subscriptions)
```

just-software-dev (approved):

PR #505, opened 2024-07-26
[trello.com/c/Kv0L5rYH] Feat: Added read/unread quick action

IanaaDvlp:


Adamant/Models/CoreData/Chatroom+CoreDataClass.swift
```diff
@@ -23,6 +23,11 @@ public class Chatroom: NSManagedObject {
         lastTransaction?.isUnread = false
     }
     
+    func markAsUnread() {
+        hasUnreadMessages = true
+        lastTransaction?.isUnread = true
+    }
```
just-software-dev, line 29: Why don't we use the exactly same algorithm as in `func markAsReaded()` but the opposite?

Adamant/Modules/ChatsList/ChatListViewController.swift
```diff
@@ -1074,15 +1084,18 @@ extension ChatListViewController {
     private func makeMarkAsReadContextualAction(for chatroom: Chatroom) -> UIContextualAction {
         let markAsRead = UIContextualAction(
             style: .normal,
-            title: nil
+            title: "👀"
         ) { (_, _, completionHandler) in
-            chatroom.markAsReaded()
+            if chatroom.hasUnreadMessages || (chatroom.lastTransaction?.isUnread ?? false) {
```
just-software-dev, line 1089: Move it to a calculated property in Chatroom

just-software-dev (approved):

Adamant/Models/CoreData/Chatroom+CoreDataClass.swift
IanaaDvlp: Discussed it on the call

PR #513, opened 2024-08-02
[trello.com/c/Ut31LisJ] Fixed: Ethereum "QR does not contain address", added "amount" for ADM legacy

IanaaDvlp:


Adamant/Utilities/AdamantUriTools.swift
```diff
@@ -93,6 +100,8 @@ final class AdamantUriTools {
                     components.queryItems?.append(.init(name: "label", value: value))
                 case .message(let value):
                     components.queryItems?.append(.init(name: "message", value: value))
+                case .amount(let value):
+                    components.queryItems?.append(.init(name: "message", value: String(value)))
```
adamant-al, line 104:
```suggestion
                    components.queryItems?.append(.init(name: "amount", value: String(value)))
```

just-software-dev (approved):

PR #547, opened 2024-09-17
[trello.com/c/1WmhFW3j] Show time instead of the 'Today' date in the Chat List

IanaaDvlp:


just-software-dev (changes requested):
commented accidentally

just-software-dev (approved):

PR #548, opened 2024-09-18
[trello.com/c/bSmPAIM2] Showed fixed date only while dragging

IanaaDvlp:


just-software-dev:
The solution doesn't work. If I scroll again immediately after finishing scrolling, the header will disappear even though I am scrolling.

And the name `timerCancellable` does not represent the meaning of the property

just-software-dev (changes requested):
...

just-software-dev:
Still a bug, but a less noticeable one:
1. Start scrolling
2. Stop scrolling and start scrolling again
3. Pause scrolling (don't scroll but keep fingers on the screen)

The header hides even though dragging is not finished yet.

On the first scroll, if we scroll and then pause scrolling, the header will not disappear. So it's definitely an unclear behavior which we have to fix

PR #567, opened 2024-10-17
[trello.com/c/sI4VXuJa] UI: restyle notification view

IanaaDvlp:


PopupKit/Sources/PopupKit/Implementation/Views/NotificationView.swift
```diff
@@ -26,13 +26,12 @@ struct NotificationView: View {
         }
         .padding([.leading, .trailing], 15)
         .padding([.top, .bottom], 10)
-        .background(GeometryReader(content: processGeometry))
```
just-software-dev, line 29: Are you sure it has to be removed?

IanaaDvlp: For that, there's `gesture(dragGesture)` 
I tested it, manual removal works

just-software-dev (approved):

PR #584, opened 2024-11-06
[trello.com/c/QCnKo03A] Fixed: adjust timestamp to avoid future date error

IanaaDvlp:


just-software-dev (approved):

adamant-al:
Outdated, replaced with Bug: Unable to send a message because of timestamp is in future (copied from PWA) https://trello.com/c/QCnKo03A

PR #629, opened 2024-12-30
[trello.com/c/ysg4LpCL] Balance vibro fix

just-software-dev:


Adamant/Modules/Wallets/Bitcoin/BtcWalletService.swift
```diff
@@ -266,6 +266,7 @@ final class BtcWalletService: WalletCoreProtocol, @unchecked Sendable {
         }
     }
     
+    @MainActor
```
just-software-dev, line 269: @MainActor to synchronise the method. As it could be called often, there is broken behavior if it's working in parallel

We could use any global actor here as well, but it would not affect app's speed as the operations are very light

ChrisBenua (approved):
lgtm

PR #631, opened 2025-01-02
[trello.com/c/57u3CHwN] Disabled keyboard extensions within the app

art-divin:
## Description

This MR disables external keyboard support within the app.

## Context

Due to the fact that external keyboards may send the data over to proprietary servers, it is reasonable to disable the support for keyboard extensions (iOS).

## Assets

|Before|After|
:-:|:-:
before|after

### DIFF
--- a/Adamant/App/AppDelegate.swift
+++ b/Adamant/App/AppDelegate.swift
@@ -331,7 +331,7 @@ class AppDelegate: UIResponder, UIApplicationDelegate {
     }
   
     func application(
-        _ application: UIApplication,
+        _: UIApplication,
         shouldAllowExtensionPointIdentifier extensionPointIdentifier: UIApplication.ExtensionPointIdentifier
     ) -> Bool {
         switch extensionPointIdentifier {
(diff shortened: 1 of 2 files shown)


BitcoinKit/Package.swift
```diff
@@ -11,7 +11,7 @@ let package = Package(
     ],
     dependencies: [
         .package(name: "OpenSSL", url: "https://github.com/krzyzanowskim/OpenSSL.git", .upToNextMinor(from: "1.1.180")),
-        .package(name: "Web3swift", url: "https://github.com/skywinder/web3swift.git", .upToNextMajor(from: "3.0.0")) //,
+        .package(name: "Web3swift", url: "https://github.com/skywinder/web3swift.git", .exact("3.2.2"))
```
just-software-dev, line 14: Will be fixed in a separate PR

just-software-dev (changes requested):

Adamant/App/AppDelegate.swift
```diff
@@ -329,6 +329,18 @@ class AppDelegate: UIResponder, UIApplicationDelegate {
         
         repeater.resumeAll()
     }
+  
+    func application(
+        _ application: UIApplication,
```
just-software-dev, line 334: Minor:
`_: UIApplication,`

just-software-dev (approved):

PR #634, opened 2025-01-08
[trello.com/c/ciHCs1De]: fix not updating row height after balance change

ChrisBenua:
`.cellUpdate` from 141 line was overwritten in 158 line. So, code from line 141 is useless, we can delete it. But it was not the only reason caused bug.

Now height of balance cell is calculated based on current row ViewModel: `$0.cell.height = { row.value?.fiat != nil ? BalanceTableViewCell.fullHeight : BalanceTableViewCell.compactHeight }`.

According to docs, to update cell height (not the content), we should call `updateCell` just before `reload` call. Without `reload` call height of cell will not be updated

### DIFF
--- a/Adamant/Modules/Wallets/WalletViewControllerBase.swift
+++ b/Adamant/Modules/Wallets/WalletViewControllerBase.swift
@@ -156,7 +156,6 @@ class WalletViewControllerBase: FormViewController, WalletViewController {
         
         balanceRow.cell.selectionStyle = .gray
         balanceRow.cellUpdate { (cell, _) in
-            cell.accessoryType = .disclosureIndicator
             cell.titleLabel.text = BaseRows.balance.localized
         }.onCellSelection { [weak self] (_, _) in
             guard
--- a/Adamant/Modules/Wallets/WalletViewControllerBase.swift
+++ b/Adamant/Modules/Wallets/WalletViewControllerBase.swift
@@ -120,6 +120,7 @@ class WalletViewControllerBase: FormViewController, WalletViewController {
         
         // MARK: Balance
         let balanceRow = BalanceRow { [weak self] in
+            $0.cell.accessoryType = .disclosureIndicator
             $0.tag = BaseRows.balance.tag
             $0.cell.titleLabel.text = BaseRows.balance.localized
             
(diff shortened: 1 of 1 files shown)


Adamant/Modules/Wallets/WalletViewControllerBase.swift
```diff
@@ -134,31 +135,13 @@ class WalletViewControllerBase: FormViewController, WalletViewController {
...
-            cell.titleLabel.text = BaseRows.balance.localized
+
+            let row = $0
+            $0.cell.height = { row.value?.fiat != nil ? BalanceTableViewCell.fullHeight : BalanceTableViewCell.compactHeight }
         }
         
         balanceRow.cell.selectionStyle = .gray
-        balanceRow.cellUpdate { (cell, _) in
-            cell.accessoryType = .disclosureIndicator
-            cell.titleLabel.text = BaseRows.balance.localized
```
ChrisBenua, line 160: Вот это я зря убрал. При изменении языка не обновится titleLabel ячейки

Adamant/Modules/Wallets/WalletViewControllerBase.swift
```diff
@@ -134,29 +135,13 @@ class WalletViewControllerBase: FormViewController, WalletViewController {
...
-                balance: self?.service?.core.wallet?.balance ?? 0,
-                symbol: symbol,
-                alert: self?.service?.core.wallet?.notifications,
-                isBalanceInitialized: self?.service?.core.wallet?.isBalanceInitialized ?? false
-            )
-            
-            let height = row.value?.fiat != nil ? BalanceTableViewCell.fullHeight : BalanceTableViewCell.compactHeight
-            
-            cell.height = { height }
-            cell.titleLabel.text = BaseRows.balance.localized
```
just-software-dev, line 154: Why so much of code is removed?

ChrisBenua: It was dead code, it never executes. Because line 158 overwrite cellUpdateCallback variable inside Eureka

just-software-dev (approved):

PR #641, opened 2025-01-11
[trello.com/c/AmlmMsUi] Network layer fixes

just-software-dev:
1. Fixed extra ERC20 balance requests. All invisible wallets were requesting their balances just like the visible wallets.
2. It seems sometimes network requests still fail on entering background or returning to the active state. So I forcefully kill health check requests on entering background now to avoid false statuses.


just-software-dev:
I'm merging it as a hotfix. If something is wrong, I will revert the Pull Request

PR #648, opened 2025-01-15
trello.com/c/Uh04lfmQ default "fastest node" button value

leonid174:
By default, all items except .amd should have a value of true. The current fix adjusts the default logic to ensure it functions correctly.

### DIFF
--- a/Adamant/Modules/CoinsNodesList/ViewModel/CoinsNodesListViewModel.swift
+++ b/Adamant/Modules/CoinsNodesList/ViewModel/CoinsNodesListViewModel.swift
@@ -47,6 +47,8 @@ final class CoinsNodesListViewModel: ObservableObject {
 
 private extension CoinsNodesListViewModel {
     func setup() {
+        setInitialFastestModeState()
+
         state.sections = processedGroups.compactMap {
             guard let info = apiServiceCompose.get($0)?.nodesInfo else { return nil }
             return mapper.map(group: $0, nodesInfo: info)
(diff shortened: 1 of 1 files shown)


ChrisBenua:
Please add PR description

Adamant/Modules/CoinsNodesList/ViewModel/CoinsNodesListState.swift
```diff
@@ -17,7 +17,7 @@ struct CoinsNodesListState: Equatable {
     static var `default`: Self {
         Self(
             sections: .init(),
-            fastestNodeMode: false,
+            fastestNodeMode: true,
```
just-software-dev, line 20: But it's just a default value of the state to be represented 🧐
I am not sure if the problem is actually solved

ChrisBenua: seems to me that changed should be done here:

```
public extension NodeGroup {
    var defaultFastestNodeMode: Bool {
        switch self {
        case .adm:
            return false
        case .eth, .doge, .dash, .btc, .klyNode, .klyService, .ipfs, .infoService:
            return true
        }
    }
}
```

crypt-eng: It's the default value of a state, the fix is definitely has to be in some other place

just-software-dev (approved):

vovameister (approved):

ChrisBenua (approved):
lgtm

PR #649, opened 2025-01-15
[trello.com/c/etjLStc6] KVS-related code refactoring

just-software-dev:
**Before:**
A blockchain address is passed as a parameter, the keypair is being received just before sending the address to the KVS. It can cause inconsistency and bugs such as saving a blockchain address from an old account to a new account's storage.

**After:**
A model, containing all the required data, is generated on wallet initialization. Then it's passed to methods to store data in KVS. No more inconsistency.


ChrisBenua (approved):
lgtm

PR #653, opened 2025-01-17
[trello.com/c/QCnKo03A] Changed time on messages logic

PingPongEzZ:
Revert previous commit due to wrong implementation of time correction of a transaction


ChrisBenua (approved):

Adamant/Modules/Chat/ViewModel/ChatViewModel.swift
```diff
@@ -1425,7 +1425,9 @@ private extension ChatViewModel {
                 return
             }
         case .serverError:
-            dialog.send(.richError(error))
+            if error.localizedDescription.contains("Timestamp is in the future") {
+                dialog.send(.error(.adamant.alert.timeAheadError, supportEmail: false))
+            }
```
crypt-eng, line 1430: Let's add an enum with possible server errors. `timestampInTheFuture` and `unknown(String)` for now. Make this enum an associated value for the `serverError`

Add mapping to `ServerResponse+Resolver.swift`

P. S. `accountNotFound` error ideally has to be in this enum as well, but we don't need more refactoring, so let's keep it as it is now

crypt-eng: Update: just add another case to the `ChatsProviderError` enum

just-software-dev (approved):

PR #657, opened 2025-01-20
[trello.com/c/qDelB45V] Add the height and the version to IPFS

PingPongEzZ:
Added mapping of timestamp to a height and version of IPFS

### DIFF
--- a/Adamant/Models/IPFSNodeStatus.swift
+++ b/Adamant/Models/IPFSNodeStatus.swift
@@ -10,37 +10,8 @@ import Foundation
 import CommonKit
 
 struct IPFSNodeStatus: Decodable {
-    private enum CodingKeys: String, CodingKey {
-        case version
-        case timestamp
-    }
-    
-    let height: Int
-    let version: Version?
-    
-    init(from decoder: any Decoder) throws {
-        let container = try decoder.container(keyedBy: CodingKeys.self)
-        let versionStringValue = try container.decodeIfPresent(String.self, forKey: .version)
-        if let versions = versionStringValue?.components(separatedBy: ".").compactMap({ Int($0) }) {
-            version = .init(versions)
-        } else {
-            version = nil
-        }
-        let timestamp = try container.decode(UInt64.self, forKey: .timestamp)
-        let timeStampInSeconds = String(timestamp / 1000)
-        
-        /// Chicking the len of string representation of a timestamp in seconds
-        /// to cut first two symbols from it and convert it to height
-        if timeStampInSeconds.count >= 2 {
-            let subString = timeStampInSeconds[
-                timeStampInSeconds.index(timeStampInSeconds.startIndex, offsetBy: 2)..<timeStampInSeconds.endIndex
-            ]
-            let string = String(subString)
-            height = Int(string) ?? .zero
-        } else {
-            height = .zero
-        }
-    }
+    let timestamp: UInt64
+    let version: String
 }
 
 /* JSON
--- a/Adamant/Models/IPFSNodeStatus.swift
+++ b/Adamant/Models/IPFSNodeStatus.swift
@@ -7,9 +7,33 @@
 //
 
 import Foundation
+import CommonKit
 
-struct IPFSNodeStatus: Codable {
-    let version: String
+struct IPFSNodeStatus: Decodable {
+    private enum CodingKeys: String, CodingKey {
+        case version
+        case timestamp
+    }
+    
+    let height: Int
+    let version: Version?
+    
+    init(from decoder: any Decoder) throws {
+        let container = try decoder.container(keyedBy: CodingKeys.self)
+        let versionStringValue = try container.decodeIfPresent(String.self, forKey: .version)
+        if let versions = versionStringValue?.components(separatedBy: ".").compactMap({ Int($0) }) {
+            version = .init(versions)
+        } else {
+            version = nil
+        }
+        let timestamp = try container.decode(UInt64.self, forKey: .timestamp)
+        let timeStampInSeconds = String(timestamp / 1000)
+        let subString = timeStampInSeconds[
+            timeStampInSeconds.index(timeStampInSeconds.startIndex, offsetBy: 2)..<timeStampInSeconds.endIndex
+        ]
+        let string = String(subString)
+        height = Int(string) ?? .zero
+    }
 }
 
 /* JSON
--- a/Adamant/Models/IPFSNodeStatus.swift
+++ b/Adamant/Models/IPFSNodeStatus.swift
@@ -28,11 +28,18 @@ struct IPFSNodeStatus: Decodable {
         }
         let timestamp = try container.decode(UInt64.self, forKey: .timestamp)
         let timeStampInSeconds = String(timestamp / 1000)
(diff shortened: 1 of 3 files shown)


Adamant/Models/IPFSNodeStatus.swift
```diff
@@ -7,9 +7,33 @@
...
+        let versionStringValue = try container.decodeIfPresent(String.self, forKey: .version)
+        if let versions = versionStringValue?.components(separatedBy: ".").compactMap({ Int($0) }) {
+            version = .init(versions)
+        } else {
+            version = nil
+        }
+        let timestamp = try container.decode(UInt64.self, forKey: .timestamp)
+        let timeStampInSeconds = String(timestamp / 1000)
+        let subString = timeStampInSeconds[
+            timeStampInSeconds.index(timeStampInSeconds.startIndex, offsetBy: 2)..<timeStampInSeconds.endIndex
```
ChrisBenua, line 32: You can easily have crash here if startIndex + 2 < endIndex. Please add guard check

ChrisBenua:
Please add PR description

Adamant/Models/IPFSNodeStatus.swift
PingPongEzZ: Not easily, closer to never, but yeah, just for safety it should be added

PingPongEzZ:
> Please add PR description

Is there any reasons to duplicate a descriptions if we pin task to a PR

ChrisBenua (approved):
lgtm

Adamant/Models/IPFSNodeStatus.swift
```diff
@@ -7,9 +7,40 @@
...
+        /// to cut first two symbols from it and convert it to height
+        if timeStampInSeconds.count >= 2 {
+            let subString = timeStampInSeconds[
+                timeStampInSeconds.index(timeStampInSeconds.startIndex, offsetBy: 2)..<timeStampInSeconds.endIndex
+            ]
+            let string = String(subString)
+            height = Int(string) ?? .zero
+        } else {
+            height = .zero
+        }
```
crypt-eng: Don't use custom decoding. Use DTOs (even though custom decoding is used in legacy)

- DTO (Date Transfer Object) — the representation of the JSON without any app-related logics.
- Model — the model that is generated out of the DTO.

Check out `NodesKeychainDTO`

crypt-eng: In this simple case you don't even need a model, you can use DTO only

just-software-dev (changes requested):
...

Adamant/Models/IPFSNodeStatus.swift
PingPongEzZ: I think there is no need to use DTO in our case because this model are already used to create a DTO from itself in `func getStatusInfo(origin: NodeOrigin) async -> ApiServiceResult<NodeStatusInfo>` and maps into a NodeStatusInfo. I've made this mapping in decoding to prevent these logics in the function above to prevent extra logic in it.

Adamant/Services/FilesNetworkManager/IPFSApiCore.swift
```diff
@@ -34,14 +34,34 @@ extension IPFSApiCore: BlockchainHealthCheckableService {
...
                 wsEnabled: false,
                 wsPort: nil,
-                version: nil
+                version: getVersionFrom(stringVersion: data.version)
             )
         }
     }
+    
+    private func getHeightFrom(timestamp: UInt64) -> Int {
+        let timeStampInSeconds = String(timestamp / 1000)
```
Lainaaa: Is not "timestamp" one word? 
if one word - we need to make camel case in a right way.
If not - we need to make other properties to follow your case

PingPongEzZ: fixed

just-software-dev (approved):

Adamant/Services/FilesNetworkManager/IPFSApiCore.swift
```diff
@@ -34,14 +34,34 @@ extension IPFSApiCore: BlockchainHealthCheckableService {
...
-                version: nil
+                version: getVersionFrom(stringVersion: $0.version)
             )
         }
     }
+    
+    private func getHeightFrom(timestamp: UInt64) -> Int {
+        let timestampInSeconds = String(timestamp / 1000)
+        
+        /// Chicking the len of string representation of a timestamp in seconds
```
crypt-eng: `Checking`, not `Chicking`

Adamant/Services/FilesNetworkManager/IPFSApiCore.swift
```diff
@@ -34,14 +34,34 @@ extension IPFSApiCore: BlockchainHealthCheckableService {
...
+            return Int(string) ?? .zero
+        } else {
+            return .zero
+        }
+    }
+    
+    private func getVersionFrom(stringVersion: String) -> Version? {
+        .init(stringVersion)
+    }
 }
```
crypt-eng, line 54: We use `private extension` in the project for defining private methods. There will be no need to write `private` in every function

Adamant/Services/FilesNetworkManager/IPFSApiCore.swift
```diff
@@ -34,14 +34,34 @@ extension IPFSApiCore: BlockchainHealthCheckableService {
...
+            let string = String(subString)
+            return Int(string) ?? .zero
+        } else {
+            return .zero
+        }
+    }
+    
+    private func getVersionFrom(stringVersion: String) -> Version? {
+        .init(stringVersion)
+    }
```
crypt-eng, line 53: This function does nothing, it's better to remove it

Adamant/Models/IPFSNodeStatus.swift
```diff
@@ -7,8 +7,10 @@
 //
 
 import Foundation
+import CommonKit
```
crypt-eng: There is no need to import `CommonKit` in this file

Adamant/Services/FilesNetworkManager/IPFSApiCore.swift
```diff
@@ -34,14 +34,34 @@ extension IPFSApiCore: BlockchainHealthCheckableService {
...
+        if timestampInSeconds.count >= 2 {
+            let subString = timestampInSeconds[
+                timestampInSeconds.index(timestampInSeconds.startIndex, offsetBy: 2)..<timestampInSeconds.endIndex
+            ]
+            let string = String(subString)
+            return Int(string) ?? .zero
+        } else {
+            return .zero
+        }
+    }
```
crypt-eng: The entire function is overcomplicated. Use math: `number / 1_000 % 100_000_000`

just-software-dev (changes requested):
...

Adamant/Services/FilesNetworkManager/IPFSApiCore.swift
PingPongEzZ: Yeah that's better abs

just-software-dev (approved):

PR #658, opened 2025-01-20
[trello.com/c/dmJg0inY]: DogeWalletService tests

ChrisBenua:
Initially open PR to EthWalletService tests branch to minimise diff.

Used only two hacks: 
1. `DogeWallet` raw initializer (marked it as deprecated with message "For testing purposes only" and made it available only in DEBUG builds)
2. `DogeWalletService.setWalletForTests` - method to set wallet easily for testing purposes, (marked it as deprecated with message "For testing purposes only"  and made it available only in DEBUG builds)

In tests I caught every custom error thrown by `createTransaction` method, verified that parameters passed to `BitcoinKit.Transaction.createTransactions` are expected.

PR is very similar to BtcWalletServiceTest but has some unique caveats to workaround: 
1. To use mocks for `DogeApiService` protocol `DogeApiServiceProtocol` was introduced, but I had to slightly change internals of `DogeApiService` to match protocol requirements (public interface didn't change)
2. `DogeApiServiceProtocolMock` redirect all calls to `_api` property, also `api` property (introduced to match protocol requirement) redirects all calls to `_api`, so whenever `DogeWalletService` calls network request using either `dogeApiService.request` or `dogeApiService.api.request` - we don't need extra mocks for each case

### DIFF
--- a/Adamant/Modules/Wallets/Doge/DogeApiService.swift
+++ b/Adamant/Modules/Wallets/Doge/DogeApiService.swift
@@ -45,19 +45,23 @@ final class DogeApiCore: BlockchainHealthCheckableService, Sendable {
     }
 }
 
-final class DogeApiService: ApiServiceProtocol {
-    let api: BlockchainHealthCheckWrapper<DogeApiCore>
+final class DogeApiService: DogeApiServiceProtocol {
+    var api: DogeInternalApiProtocol {
+        _api
+    }
+    
+    let _api: BlockchainHealthCheckWrapper<DogeApiCore>
     
     @MainActor
-    var nodesInfoPublisher: AnyObservable<NodesListInfo> { api.nodesInfoPublisher }
+    var nodesInfoPublisher: AnyObservable<NodesListInfo> { _api.nodesInfoPublisher }
     
     @MainActor
-    var nodesInfo: NodesListInfo { api.nodesInfo }
+    var nodesInfo: NodesListInfo { _api.nodesInfo }
     
-    func healthCheck() { api.healthCheck() }
+    func healthCheck() { _api.healthCheck() }
     
     init(api: BlockchainHealthCheckWrapper<DogeApiCore>) {
-        self.api = api
+        self._api = api
     }
     
     func request<Output>(
@@ -75,3 +79,5 @@ final class DogeApiService: ApiServiceProtocol {
         }
     }
 }
+
+extension BlockchainHealthCheckWrapper: DogeInternalApiProtocol where Service == DogeApiCore {}
--- a/Adamant/Modules/Wallets/Doge/DogeWalletService.swift
+++ b/Adamant/Modules/Wallets/Doge/DogeWalletService.swift
@@ -55,7 +55,8 @@ final class DogeWalletService: WalletCoreProtocol, @unchecked Sendable {
     
     // MARK: - Dependencies
     var apiService: AdamantApiServiceProtocol!
-    var dogeApiService: DogeApiService!
+    var dogeApiService: DogeApiServiceProtocol!
+    var btcTransactionFactory: BitcoinKitTransactionFactoryProtocol!
     var accountService: AccountService!
     var dialogService: DialogService!
     var addressConverter: AddressConverter!
@@ -390,6 +391,7 @@ extension DogeWalletService: SwinjectDependentService {
         accountService = container.resolve(AccountService.self)
         apiService = container.resolve(AdamantApiServiceProtocol.self)
         dialogService = container.resolve(DialogService.self)
+        btcTransactionFactory = container.resolve(BitcoinKitTransactionFactoryProtocol.self)
         addressConverter = container.resolve(AddressConverterFactory.self)?
             .make(network: network)
         dogeApiService = container.resolve(DogeApiService.self)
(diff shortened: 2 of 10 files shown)


Adamant/Modules/Wallets/Doge/DogeWalletService.swift
```diff
@@ -682,6 +684,15 @@ extension DogeWalletService {
     }
 }
 
+#if DEBUG
+extension DogeWalletService {
```
PingPongEzZ, line 695: If you're using it in tests only u probably may create an function in the tests target. And there would be no need to add #if #else statement and @available statement. Just a nit

Adamant/Modules/Wallets/Doge/DogeApiService.swift
```diff
@@ -75,3 +79,5 @@ final class DogeApiService: ApiServiceProtocol {
         }
     }
 }
+
+extension BlockchainHealthCheckWrapper: DogeInternalApiProtocol where Service == DogeApiCore {}
```
PingPongEzZ, line 83: I'm not sure. But this looks like a wrong thing. `BlockchainHealthCheckWrapper` uses a `BlockchainHealthCheckableService` as a generic. And here in this file for `DogeApiService` you are changing the contract of it.
If you want to mock `DogeApiService` instance it might be correct to keep `ApiServiceProtocol` conformance and mock it as it is. Maybe we could have a sync call because it's hard to understand the usage of this code(the code is fine just need to understand why this code has been written as it is).
Or as you did in the BTC tests you've made a new protocol inherited from `ApiServiceProtocol` and mocked it.

PingPongEzZ: Looks like you've made this just for the tests to prevent `api` mocks from ApiServiceProtocol. I think that's not the best because it makes the code a bit complex and harder to understand why this is written just like this while the others ApiServices works without any proxies for testing.

PingPongEzZ (approved):

Adamant/Modules/Wallets/Doge/DogeWalletService.swift
ChrisBenua: It's impossible to assign private(set) properties outside class definition file

Adamant/Modules/Wallets/Doge/DogeApiService.swift
ChrisBenua: There is no way I can get around this code `_ = try await dogeApiService.api.request(waitsForConnectivity: false)` in `sendTransaction` method without covering `BlockchainHealthCheckWrapper` with protocol for mocking. Or try to create `BlockchainHealthCheckWrapper` and mock its dependencies, but it will cause event more protocols 'just for tests' and mocks

In this specific case we can't use `try await dogeApiService.request(...)` method because of custom error handling. In all other cases in DogeWalletService  `try await dogeApiService.request(...)` is used.

In BTCWalletServiceTest there were no calls like `_ = try await btcApiService.api.request(waitsForConnectivity: false)`, only calls ``_ = try await btcApiService.request(waitsForConnectivity: false)`, so there was no need for additional mocking

Adamant/Modules/Wallets/Doge/DogeWalletService.swift
PingPongEzZ: Awh, there is private(set) usage in here. Haven't c it

Adamant/Modules/Wallets/Doge/DogeApiService.swift
PingPongEzZ: Maybe if there will be more times when you need to mock `ApiServiceProtocol` would be good to make a mock impl of `BlockchainHealthCheckWrapper`. IDK. It's fine but I just want to understand what's happening here because some code might seem not so obvious from the other look

ChrisBenua: I'll take a look on how to mock BlockchainHealthCheckWrapper. If it will be too complicated compared to this mock - I will post comment here

ChrisBenua: In BlockchainHealthCheckWrapper and its superclass there are too much unmockable code like Timers, async Tasks etc. It's much simpler to stay with this kind of mock

AdamantTests/Modules/Wallets/DogeWalletServiceTests.swift
```diff
@@ -0,0 +1,307 @@
...
+            .make(network: DogeMainnet())
+            .convert(address: address)
+    }
+    
+    private func assertAddressesEqual(_ lhs: Address, _ rhs: Address, file: StaticString = #file, line: UInt = #line) {
+        XCTAssertEqual(lhs.lockingScript, rhs.lockingScript, file: file, line: line)
+        XCTAssertEqual(lhs.stringValue, rhs.stringValue, file: file, line: line)
+        XCTAssertEqual(lhs.lockingScriptPayload, rhs.lockingScriptPayload, file: file, line: line)
+        XCTAssertEqual(lhs.scriptType, rhs.scriptType, file: file, line: line)
+    }
```
crypt-eng, line 238: Use `private extension` for private methods

just-software-dev (approved):

PR #665, opened 2025-01-22
[trello.com/c/kie7Tdnq] Create service to manage all alive secret wallets 

Lainaaa:
part of https://trello.com/c/kie7Tdnq/803-dev-create-service-to-manage-all-alive-secret-wallets

Prev:
WalletServiceCompose
                                     |                                              
                                    v                                                
Implementation:   AdamantWalletServiceCompose

Curr:
                                        WalletServiceCompose
                                       /                                        \
                                      v                                         v
   PublicWalletServiceCompose                           SecretWalletServiceCompose
                                          /                                                                                     \
                                         v                                                                                     v
Implementation:  AdamantPublicWalletServiceCompose               AdamantSecretWalletServiceCompose

This structure is better because it adds a layer of abstraction, allowing clear separation of responsibilities:
PublicWalletServiceCompose: For cases needing only public wallets.
SecretWalletServiceCompose: For cases needing only secret wallets.
WalletServiceCompose: For scenarios needing both.
This improves flexibility, enforces clearer boundaries, and makes the system easier to maintain and extend.


Adamant/Modules/Account/AccountViewController/AccountViewController.swift
```diff
@@ -67,6 +67,10 @@ final class AccountViewController: FormViewController {
     
     private var notificationsSet: Set<AnyCancellable> = []
     
+    private var currentWallet: WalletServiceCompose {
```
ChrisBenua, line 70: Are you sure with naming? currentWallet looks like entity, not service. just a nit

Adamant/Modules/Wallets/SecretWalletsService/SecretWalletsService.swift
```diff
@@ -0,0 +1,25 @@
...
+//  Adamant
+//
+//  Created by Dmitrij Meidus on 22.01.25.
+//  Copyright © 2025 Adamant. All rights reserved.
+//
+
+protocol SecretWalletsService {
+    var publicWallet: PublicWalletServiceCompose { get }
+    var secretWallets: [String: SecretWalletServiceCompose] { get }
+    var currentWallet: PublicWalletServiceCompose { get }
```
ChrisBenua, line 12: what's the difference between publicWallet and currentWallet? in init below seems like they are the same

Podfile.lock
```diff
@@ -21,4 +21,4 @@ SPEC CHECKSUMS:
 
 PODFILE CHECKSUM: a30619b79caa4b5a7497b0600d449f34b5620eec
 
-COCOAPODS: 1.15.2
+COCOAPODS: 1.16.2
```
ChrisBenua, line 24: discard this diff please. May be we should start to use bundler to make sure all of us have the same pods version

PR #666, opened 2025-01-22
[trello.com/c/QSbUdocE]: KlyWalletService tests

ChrisBenua:
KlyWalletService test.

For testing purposes `KlyNodeApiServiceProtocol` and `KlyTransactionFactoryProtocol` were introduced.

For testing URL-request inside `sendTransaction` `RpcRequestBody` was used. To return real response body, I've used `KlyTransactionSubmitModel`.

Because LiskKit inside uses private `URLSession`, default practice using `MockURLProtocol` won't work, because it requires to instantiate `URLSession` with custom `URLSessionConfiguration`. So I've used swizzling for mocking `URLSession.init`. Because internally `URLSession.init` is actually `class method`, so for swizzling we should exchange two class methods implementations:

```
guard let originalMethod = class_getClassMethod(URLSession.self, originalSelector),
           let swizzledMethod = class_getClassMethod(URLSession.self, swizzledSelector) else { return }
```

Inside swizzled method we change actual `URLConfiguration` with custom `URLConfiguration` with mocking using `MockURLProtocol`

### DIFF
--- a/AdamantTests/Modules/Wallets/KlyWalletServiceTests.swift
+++ b/AdamantTests/Modules/Wallets/KlyWalletServiceTests.swift
@@ -130,18 +130,22 @@ final class KlyWalletServiceTests: XCTestCase {
         XCTAssertNil(result2.error)
         XCTAssertTrue(calledCompletion)
     }
-    
-    private static func applyURLSessionSwizzling() {
+}
+
+// MARK: Private
+
+private extension KlyWalletServiceTests {
+    static func applyURLSessionSwizzling() {
         URLSession.swizzleInitializer()
     }
     
-    private static func makeSessionConfig() -> URLSessionConfiguration {
+    static func makeSessionConfig() -> URLSessionConfiguration {
         let config = URLSessionConfiguration.ephemeral
         config.protocolClasses = [MockURLProtocol.self]
         return config
     }
     
-    private func makeKlySendMock(expectedHash: String, _ onCall: @escaping () -> Void) {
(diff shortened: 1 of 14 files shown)


AdamantTests/Modules/Wallets/KlyWalletServiceTests.swift
```diff
@@ -0,0 +1,232 @@
...
+        URLSessionSwizzlingHolder._stubbedUrlSessionConfiguration = nil
+        super.tearDown()
+    }
+    
+    func test_createTransaction_noWalletServiceThrowsError() async throws {
+        // given
+        sut.setWalletForTests(nil)
+        
+        // when
+        let result = await Swift.Result(catchingAsync: {
```
ChrisBenua, line 50: Have to use Swift.Result here because LiskKit has Result type

AdamantTests/Modules/Wallets/KlyWalletServiceTests.swift
```diff
@@ -0,0 +1,267 @@
...
+        // when 2
+        XCTAssertNil(result2.error)
+        XCTAssertTrue(calledCompletion)
+    }
+    
+    private static func applyURLSessionSwizzling() {
+        URLSession.swizzleInitializer()
+    }
+    
+    private static func makeSessionConfig() -> URLSessionConfiguration {
```
crypt-eng: Use `private extension` for private methods

just-software-dev (approved):

vovameister (approved):

PR #671, opened 2025-01-24
[trello.com/c/Htm0a2Qp]: DashWalletService tests

ChrisBenua:
DashWalletService - is like `BtcWalletService` but with slightly different API contract + logic with lastTransactionId.

For testing logic with lastTransactionId I've made `DashLastTransactionStorageProtocol`. It's used in both unit-tests and integration tests. 

Test where lastTransaction has not enough confirmations is named `test_createTransaction_throwsErrorWhenHasLastTransactionIdAndNotEnoughConfirmations`.

All other cases are almost similar to `BtcWalletService` tests with slightly mock differences

### DIFF
--- a/Adamant/App/DI/AppAssembly.swift
+++ b/Adamant/App/DI/AppAssembly.swift
@@ -202,6 +202,12 @@ struct AppAssembly: MainThreadAssembly {
             ))
         }.inObjectScope(.container)
         
+        
+        // MARK: DashLastTransactionStorage
+        container.register(DashLastTransactionStorageProtocol.self) { r in
+            DashLastTransactionStorage(securedStore: r.resolve(SecuredStore.self)!)
+        }.inObjectScope(.container)
+        
         // MARK: LskNodeApiService
         container.register(KlyNodeApiService.self) { r in
             KlyNodeApiService(api: .init(
--- a/Adamant/Modules/Wallets/Dash/DashWalletService.swift
+++ b/Adamant/Modules/Wallets/Dash/DashWalletService.swift
@@ -65,9 +65,10 @@ final class DashWalletService: WalletCoreProtocol, @unchecked Sendable {
     
     // MARK: - Dependencies
     var apiService: AdamantApiServiceProtocol!
-    var dashApiService: DashApiService!
+    var dashApiService: DashApiServiceProtocol!
     var accountService: AccountService!
-    var securedStore: SecuredStore!
+    var lastTransactionStorage: DashLastTransactionStorageProtocol!
+    var transactionFactory: BitcoinKitTransactionFactoryProtocol!
     var dialogService: DialogService!
     var addressConverter: AddressConverter!
     var coreDataStack: CoreDataStack!
@@ -90,32 +91,10 @@ final class DashWalletService: WalletCoreProtocol, @unchecked Sendable {
     
     var lastTransactionId: String? {
         get {
-            guard
-                let hash: String = self.securedStore.get("lastDashTransactionId"),
-                let timestampString: String = self.securedStore.get("lastDashTransactionTime"),
-                let timestamp = Double(string: timestampString)
-            else { return nil }
-            
-            let date = Date(timeIntervalSince1970: TimeInterval(timestamp))
-            let timeAgo = -1 * date.timeIntervalSinceNow
-            
-            if timeAgo > 10 * 60 { // 10m waiting for transaction complete
-                self.securedStore.remove("lastDashTransactionTime")
-                self.securedStore.remove("lastDashTransactionId")
-                return nil
-            } else {
-                return hash
-            }
+            lastTransactionStorage.getLastTransactionId()
         }
         set {
-            if let value = newValue {
-                let timestamp = Date().timeIntervalSince1970
-                self.securedStore.set("\(timestamp)", for: "lastDashTransactionTime")
-                self.securedStore.set(value, for: "lastDashTransactionId")
-            } else {
-                self.securedStore.remove("lastDashTransactionTime")
-                self.securedStore.remove("lastDashTransactionId")
-            }
+            lastTransactionStorage.setLastTransactionId(newValue)
         }
     }
     
@@ -420,7 +399,8 @@ extension DashWalletService: SwinjectDependentService {
     func injectDependencies(from container: Container) {
         accountService = container.resolve(AccountService.self)
         apiService = container.resolve(AdamantApiServiceProtocol.self)
-        securedStore = container.resolve(SecuredStore.self)
+        lastTransactionStorage = container.resolve(DashLastTransactionStorageProtocol.self)
+        transactionFactory = container.resolve(BitcoinKitTransactionFactoryProtocol.self)
         dialogService = container.resolve(DialogService.self)
         addressConverter = container.resolve(AddressConverterFactory.self)?
             .make(network: network)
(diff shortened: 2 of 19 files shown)


Adamant/Modules/Wallets/Dash/DashWalletService.swift
```diff
@@ -605,6 +585,15 @@ extension DashWalletService {
     }
 }
 
+#if DEBUG
+extension DashWalletService {
```
art-divin, line 602: Could we move this into the test target instead? Even though it is marked with DEBUG, for better organization it is better to put the code related to tests under test targets only. Would it be possible @ChrisBenua?

AdamantTests/Modules/Wallets/DashWalletServiceIntegrationTests.swift
```diff
@@ -0,0 +1,135 @@
...
+    
+    static let passphrase = "village lunch say patrol glow first hurt shiver name method dolphin dead"
+    
+    static let expectedTransactionID = "d4cf3fde45d0e7ba855db9621bdc6da091856011d86f199bebd1937f7b63020a"
+    
+    static let expectedTransactionHex = "0100000001721f0c2437124acb8a20fea5af60908057b086118b240119adcb39c890b4c3a2010000006a473044022002e19bc62748ca3f34a6e5f1aeab31bb3dc43997792d9551e9b8c51a094abbae02200e3c9bde7c60326ef6984045a81304a9915e8fbce96de01813fe89886ef26453012102cd3dcbdfc1b77e54b3a8f273310806ab56b0c2463c2f1677c7694a89a713e0d0ffffffff0240420f00000000001976a914931ef5cbdad28723ba9596de5da1145ae969a71888acb695a905000000001976a91457f6f900ac7a7e3ccab712326cd7b85638fc15a888ac00000000"
+    
+    static let unspentTranscationsData = unspentTransactionsRawJSON.data(using: .utf8)!
+    
+    static let unspentTransactionsRawJSON: String = """
```
art-divin: We might want to use actual JSON files, fixtures that is, in order to have a more unified way of editing and searching for them when writing more and more tests in future.
What do you think @ChrisBenua?

AdamantTests/Modules/Wallets/DashWalletServiceIntegrationTests.swift
```diff
@@ -0,0 +1,135 @@
...
+        lastTransactionStorageMock = DashLastTransactionStorageProtocolMock()
+        
+        sut = DashWalletService()
+        sut.lastTransactionStorage = lastTransactionStorageMock
+        sut.addressConverter = AddressConverterFactory().make(network: DashMainnet())
+        sut.dashApiService = dashApiServiceProtocolMock
+        sut.transactionFactory = BitcoinKitTransactionFactory()
+    }
+    
+    override func tearDown() {
```
art-divin, line 36: Nice touch!

AdamantTests/Modules/Wallets/DashWalletServiceIntegrationTests.swift
```diff
@@ -0,0 +1,135 @@
...
+        let result = await Result(catchingAsync: {
+            try await self.sut.createTransaction(
+                recipient: "Xp6kFbogHMD4QRBDLQdqRp5zUgzmfj1KPn",
+                amount: 0.01,
+                fee: 0.0000001,
+                comment: nil
+            )
+        })
+        
+        // then 1
```
art-divin, line 62: What about splitting this UT into 3 separate onces? 1 given/when/expect per test?
It may introduce a bit of code duplication, apart from re-using the same code for configuration put in a separate methods with arguments,
but in the long term it is:

a) easier to code review
b) easier to modify

What do you think @ChrisBenua?

AdamantTests/Modules/Wallets/DashWalletServiceTests.swift
```diff
@@ -0,0 +1,363 @@
...
+        // given
+        lastTransactionStorageMock.stubbedLastTransactionId = nil
+        
+        // when
+        let result = await Result {
+            try await self.sut.create(recipient: Constants.invalidDashAddress, amount: 10)
+        }
+        
+        // then
+        XCTAssertEqual(result.error as? WalletServiceError, .notLogged)
```
art-divin, line 94: What do you think of using `XCTAssertThrowsError` ?

AdamantTests/Stubs/DashApiServiceProtocolMock.swift
```diff
@@ -0,0 +1,40 @@
...
+//
+//  Created by Christian Benua on 23.01.2025.
+//  Copyright © 2025 Adamant. All rights reserved.
+//
+
+@testable import Adamant
+import CommonKit
+import Foundation
+
+final class DashApiServiceProtocolMock: DashApiServiceProtocol {
```
art-divin, line 13: It may be better to introduce `Sourcery`/`SwiftyMocky` or `mockolo` for mock generation, or any other pair of tools to automate this type of file creation. It will become cumbersome in the near future, especially updaing existing protocol mocks when actual protocols are modified. We can discuss it further and decide for the set of tools

Adamant/Modules/Wallets/Dash/DashWalletService.swift
ChrisBenua: Unfortunately, you can't access private properties of class outside of class declaration file. So there are only two options: make wallet property internal or add this extension for debug build only

AdamantTests/Modules/Wallets/DashWalletServiceIntegrationTests.swift
ChrisBenua: In my point of view, integration test supposed to be like that, because we're testing whole flow, from creating transaction to sending it and checking after each action that everything is ok/

Moreover, if I split it, I will have at least 20 lines of the same code in each testing methods (creating transaction). 

I'm not sure its good idea to make two testing methods totally ~55 lines long instead of 35 lines long single test function.

AdamantTests/Stubs/DashApiServiceProtocolMock.swift
ChrisBenua: Most of the times I need custom logic, which can't be autogenerated.

This mock is quite good example of custom mock logic.

And If these tests will remain the only ones in whole project (I think so), using code-generation for mocks is overkill

AdamantTests/Modules/Wallets/DashWalletServiceTests.swift
ChrisBenua: I prefer to use consistent logic for handling fail-case scenarios and success scenarios, so code looks the same

```
// fail
XCTAssertEqual(result.error as? WalletServiceError, .notLogged)
// success
XCTAssertEqual(result.value, expectedResult)

```

AdamantTests/Modules/Wallets/DashWalletServiceIntegrationTests.swift
ChrisBenua: Done

Adamant/App/DI/AppAssembly.swift
```diff
@@ -202,6 +202,12 @@ struct AppAssembly: MainThreadAssembly {
             ))
         }.inObjectScope(.container)
         
+        
+        // MARK: DashLastTransactionStorage
+        container.register(DashLastTransactionStorageProtocol.self) { r in
```
Lainaaa, line 207: Could we use 
`#if DEBUG`
in this case?

Lainaaa (approved):
lgtm

Adamant/App/DI/AppAssembly.swift
ChrisBenua: It will break injection logic inside DashWalletService

```
lastTransactionStorage = container.resolve(DashLastTransactionStorageProtocol.self)
```

So no, we can't

PingPongEzZ (approved):

AdamantTests/Stubs/DashApiServiceProtocolMock.swift
ChrisBenua: Let's agree on this for now, if something will make your eyes pop - make a ticket, I will fix it

PR #675, opened 2025-01-27
[trello.com/c/YrmN2oiR]: ERC20WalletService tests

ChrisBenua:
Implemented `ERC20WalletService` tests.

Main logic is the same as in `EthWalletService` tests, but there are specific cases where I spent most of time trying to figure out.

For example, to get `multiplier` for current token, API call is used. I mock it using `makeDecimalsMock` and its response looks like `0x0000000000000000000000000000000000000000000000000000000000000006`.

Additionally `ERC20ApiServiceProtocolMock` was needed to proxy all URL calls through custom `Web3Provider` with out custom `URLSession`

### DIFF
--- a/AdamantTests/Modules/Wallets/ERC20WalletServiceTests.swift
+++ b/AdamantTests/Modules/Wallets/ERC20WalletServiceTests.swift
@@ -159,8 +159,12 @@ final class ERC20WalletServiceTests: XCTestCase {
         XCTAssertEqual(transaction.gasPrice, BigUInt(clamping: 11).toWei())
         XCTAssertEqual(transaction.hashForSignature(), Constants.expectedHashForSignature)
     }
-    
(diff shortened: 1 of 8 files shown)


AdamantTests/Modules/Wallets/ERC20WalletServiceTests.swift
```diff
@@ -0,0 +1,258 @@
...
+        let mockData = try JSONEncoder().encode(ethResponse)
+        return (response, mockData)
+    }
+    
+    private func makeSession() -> URLSession {
+        let config = URLSessionConfiguration.ephemeral
+        config.protocolClasses = [MockURLProtocol.self]
+        return URLSession(configuration: config)
+    }
+}
```
crypt-eng, line 222: Use `private extension` for private methods

just-software-dev (approved):

Lainaaa (approved):

PR #681, opened 2025-01-30
[trello.com/c/R42vd9hV] Prevent actions if not enough balance

vovameister:
Added Prevent actions when there is not enough balance to change the chat name (contact) and to send remote push notifications. Creation of renaming alert and free token has been moved to DialogService. Free token alert change, now it have one more option (buy ADMs in the chat).


Adamant/Modules/Chat/View/Managers/ChatDialogManager.swift
```diff
@@ -251,11 +249,26 @@ private extension ChatDialogManager {
     }
     
     func showRenameAlert() {
-        guard let alert = makeRenameAlert() else { return }
+        guard let address = address else { return }
+        
+        let alert = dialogService.makeRenameAlert(
+            titleFormat:String(format: .adamant.chat.actionsBody, address),
```
Lainaaa, line 279: Space required

Adamant/Modules/ChatsList/ChatListViewController.swift
```diff
@@ -1294,46 +1297,27 @@ extension ChatListViewController {
...
+            
+            let alert = dialogService.makeRenameAlert(
+                titleFormat: String(format: .adamant.chat.actionsBody, address),
+                initialText: self.addressBook.getName(for: address),
+                needToPresent: accountService.account?.isEnoughMoneyForTransaction,
+                url: accountService.account?.address,
+                showVC: { [weak self] in
+                    self?.presentBuyAndSell()
+                }
+            ) { [weak self] newName in
```
Lainaaa, line 1301: Multiple Closures with Trailing Closure Violation

Adamant/Modules/PartnerQR/PartnerQRView.swift
```diff
@@ -25,10 +26,16 @@ struct PartnerQRView: View {
                 }
             }
         }
+        .fullScreenCover(isPresented: $viewModel.presentBuyAndSell) {
+            screenFactory.makeBuyAndSellView(action: {
```
Lainaaa, line 30: What do you think about moving it to the viewModel?

Adamant/Modules/Settings/Notifications/NotificationsView.swift
```diff
@@ -50,6 +53,11 @@ struct NotificationsView: View {
         .fullScreenCover(isPresented: $viewModel.openSafariURL) {
             SafariWebView(url: viewModel.safariURL).ignoresSafeArea()
         }
+        .fullScreenCover(isPresented: $viewModel.presentBuyAndSell) {
+            screensFactory.makeBuyAndSellView(action: {
```
Lainaaa, line 57: What do you think about moving it to the viewModel?

Adamant/Modules/Wallets/Adamant/BuyAndSellViewController.swift
```diff
@@ -153,53 +156,34 @@ final class BuyAndSellViewController: FormViewController {
...
-            return room.partner?.address == AdamantContacts.adamantExchange.address
-        })
-        
-        guard let chatroom = chatroom,
-              let chatDetail = chatDetail
-        else {
+    @MainActor
+    private func openExchangeChat() async {
+        guard let chatroom = await chatsProvider.getChatroom(for: AdamantContacts.adamantExchange.address) else {
+            print("Unable to find necessary components to present chatroom")
```
Lainaaa: Could it be better to throw error instead?

Adamant/ServiceProtocols/DialogService.swift
```diff
@@ -234,6 +234,7 @@ protocol DialogService: AnyObject {
     // MARK: - Alerts
     func showAlert(title: String?, message: String?, style: UIAlertController.Style, actions: [UIAlertAction]?, from: UIAlertController.SourceView?)
     func showAlert(title: String?, message: String?, style: AdamantAlertStyle, actions: [AdamantAlertAction]?, from: UIAlertController.SourceView?)
-    
+    func makeRenameAlert(titleFormat: String, initialText: String?, needToPresent: Bool?, url: String?, showVC: @escaping () -> Void, onRename: @escaping (String) -> Void) -> UIAlertController
```
Lainaaa: The name `needToPresent` is not really clear to me. 
Do you mean something like `shouldShowImmediately`?

Adamant/Services/AdmDialogService/AdmDialogService+Extension.swift
```diff
@@ -0,0 +1,134 @@
...
+//  AdmDialogServiceExtent.swift
+//  Adamant
+//
+//  Created by Владимир Клевцов on 25.1.25..
+//  Copyright © 2025 Adamant. All rights reserved.
+//
+import UIKit
+import SafariServices
+
+extension AdamantDialogService {
```
Lainaaa, line 11: Other extensions are already in `AdamantDialogService.swift`. 
I suggest keeping everything in one place—either in `AdamantDialogService.swift` or `AdamantDialogService+Extension.swift`.

Adamant/Modules/PartnerQR/PartnerQRView.swift
vovameister: I think the responsibility for navigation lies with the view

Adamant/ServiceProtocols/DialogService.swift
vovameister: here it is mean "need to present?" free tokens alert in the future

Adamant/Services/AdmDialogService/AdmDialogService+Extension.swift
vovameister: I think we can rename this Extension file in the future, I dont like the idea to collect all alerts in single file. We can have a 2k+ strings file in the future and its gonna be hard to read and manage well. I would suggest moving some alerts that can be grouped by features into a separate files

CommonKit/Sources/CommonKit/Models/AdamantAccount.swift
```diff
@@ -95,7 +95,11 @@ extension AdamantAccount: WrappableModel {
         )
     }
 }
-
+public extension AdamantAccount {
+    var isEnoughMoneyForTransaction: Bool {
```
ChrisBenua, line 99: isEnoughMoneyForRename?

Adamant/Services/AdmDialogService/AdmDialogService+Extension.swift
```diff
@@ -0,0 +1,134 @@
...
+            style: .default
+        ) { _ in
+            guard
+                let textField = alert.textFields?.first,
+                let newName = textField.text,
+                !newName.isEmpty
+            else { return }
+            
+            onRename(newName)
+            if !(needToPresent ?? true) {
```
ChrisBenua: I don't understand why we present alert if needToPresent == false

CommonKit/Sources/CommonKit/Models/AdamantAccount.swift
vovameister: Its not only for rename, its for message and push notification too

Adamant/Services/AdmDialogService/AdmDialogService+Extension.swift
vovameister: fixed

Adamant/Services/AdmDialogService/AdmDialogService+Extension.swift
```diff
@@ -0,0 +1,134 @@
...
+//
+import UIKit
+import SafariServices
+
+extension AdamantDialogService {
+    @MainActor
+    func makeRenameAlert(
+        titleFormat: String,
+        initialText: String?,
+                isEnoughMoney: Bool?,
```
ChrisBenua: fix indentation please

Adamant/Services/AdmDialogService/AdmDialogService+Extension.swift
```diff
@@ -0,0 +1,134 @@
...
+            guard let url = self.freeTokensURL(url: url) else { return }
+            let safari = SFSafariViewController(url: url)
+            safari.preferredControlTintColor = UIColor.adamant.primary
+            safari.modalPresentationStyle = .overFullScreen
+            
+            window.rootViewController?.present(safari, animated: true)
+        }
+        return action
+    }
+    func mackBuyTokensAction(action: @escaping () -> Void) -> UIAlertAction {
```
ChrisBenua: mackBuyTokensAction -> makeBuyTokensAction

Adamant/Services/AdmDialogService/AdmDialogService+Extension.swift
```diff
@@ -0,0 +1,134 @@
...
+        }
+    }
+    func freeTokensURL(url: String?) -> URL? {
+        guard let url = url else {
+            return nil
+        }
+        let urlString: String = .adamant.wallets.getFreeTokensUrl(for: url)
+        let tokenUrl = URL(string: urlString)
+        
+        return tokenUrl
```
ChrisBenua, line 116: Optional nit:
we can make beautiful functional chain

```
url.map { 
    .adamant.wallets.getFreeTokensUrl(for: $0) 
}.flatMap { 
    URL(string: $0)
}
```

vovameister: need to correct your code to make it compiled, I think current is fine

Adamant/Services/AdmDialogService/AdmDialogService+Extension.swift
vovameister: fixed

ChrisBenua (approved):
lgtm

Adamant/Modules/PartnerQR/PartnerQRViewModel.swift
```diff
@@ -213,44 +224,19 @@ private extension PartnerQRViewModel {
...
-        alert.modalPresentationStyle = .overFullScreen
-        return alert
-    }
     
     func makeCancelAction() -> UIAlertAction {
         .init(title: .adamant.alert.cancel, style: .cancel, handler: nil)
     }
+    func handleRename(newName: String) {
+        Task { [weak self] in
+            guard let self else { return }
```
art-divin, line 233: After guarding against weak self, there's no need to reference `self` upon use within the closure. Might be better to remove it, @just-software-dev told me this is a decision held within the project. What do you think @vovameister?

Adamant/Modules/Settings/Notifications/NotificationsViewModel.swift
```diff
@@ -24,6 +24,7 @@ final class NotificationsViewModel: ObservableObject {
     @Published var inAppSounds: Bool = false
     @Published var inAppVibrate: Bool = true
     @Published var inAppToasts: Bool = true
+    @Published var presentBuyAndSell: Bool = false
```
art-divin, line 27: In banking this type of a dialog is usually named "upsell", maybe we can rename it here to make it more aligned with this standard?
What do you think @vovameister?

Adamant/Modules/PartnerQR/PartnerQRViewModel.swift
vovameister: @art-divin if i remember correctly when we discussed this with @just-software-dev he ask me to use [weak self] for insurance even if it seems that a leak definitely cannot happen

Adamant/Modules/Settings/Notifications/NotificationsViewModel.swift
vovameister: we have BuyAndSellViewController so if we start renaming we may go too far

Adamant/ServiceProtocols/DataProviders/DataProvider.swift
```diff
@@ -9,27 +9,27 @@
...
+    var stateObserver: AnyObservable<StateEnum> { get }
     var isInitiallySynced: Bool { get }
     
     func reload() async
     func reset()
 }
 
 // MARK: - Status Equatable
-extension State: Equatable {
+extension StateEnum: Equatable {
```
art-divin: I would suggest to rename it to make the name more meaningful in the context of the DataProvider. Personally, I have renamed it to `DataProviderState` and it seems more aligned with the usages. What do you think @vovameister?

Adamant/Services/AdmDialogService/AdmDialogService+Extension.swift
```diff
@@ -0,0 +1,134 @@
...
+            style: .default
+        ) { _ in
+            guard
+                let textField = alert.textFields?.first,
+                let newName = textField.text,
+                !newName.isEmpty
+            else { return }
+            
+            onRename(newName)
+            if !(isEnoughMoney ?? true) {
```
art-divin: I am thinking, `isEnoughMoney` can be nil in case account service' `account` property is nil.

However, how can it be that the current account property is `nil` in the logged in account?
The logic in the `AccountService` seeem to be bound to the fact that the AccountService is created before the account, and so the app can function without an active account (for login etc.).

The question is, maybe we should guard the account in the account service  before presenting this dialog, and make `isEnoughMoney` non-optional?

What do you think @vovameister?

Adamant/ServiceProtocols/DataProviders/DataProvider.swift
vovameister: fixed

Adamant/Services/AdmDialogService/AdmDialogService+Extension.swift
vovameister: fixed

Adamant/Modules/PartnerQR/PartnerQRViewModel.swift
art-divin: Sure, this is not about `weak self` use, this is definitely a recommended practice.
But about using `self` keyword to reference the instance, it becomes redundant after use unwrap it explicitly using the `guard` statement.

For example, this method would look like this after my suggestion would get applied, if you agree to do so:

```swift
func handleRename(newName: String) {
        Task { [weak self] in
            guard let self else { return }
            partnerName = newName
            renameTitle = .adamant.alert.renameContact
            await addressBookService.set(name: newName, for: title)
        }
    }
```

BUT I have just checked, and most of the time we actually use `self` after unwrapping it within closures. So... Myself, I am actually OK with specifying self. Maybe I have misunderstood before about the guidance. 
From this perspective, when `self` is still specified, we can distinguish here in GitHub when reviewing code that it is referencing a member and not a temporary, local variable, for example.

Adamant/Modules/Settings/Notifications/NotificationsViewModel.swift
art-divin: Sure, makes sense 👍🏻

art-divin (approved):
Amazing work 👍🏻

PR #683, opened 2025-01-31
[trello.com/c/9E1yhlaI] Debug build speed up

vovameister:
Keystore creation is moved to an actor to avoid re-creation of the same keystore. This sped up the loading of the account on the simulator and device from ~12 seconds to ~4.5


Adamant/App/DI/AppAssembly.swift
```diff
@@ -444,5 +444,9 @@ struct AppAssembly: MainThreadAssembly {
         container.register(DefaultNodesProvider.self) { _ in
             DefaultNodesProvider()
         }.inObjectScope(.transient)
+        
+        container.register(EthBIP32ServiceProtocol.self) { r in
+            EthBIP32Service(erc20ApiService: r.resolve(ERC20ApiService.self)!)
+        }
```
ChrisBenua: Does it utilise .singleton scope? If not - can we make it singleton? Does it provide additional speed up?

Adamant/Modules/Wallets/EthBIP32Service/EthBIP32Service.swift
```diff
@@ -0,0 +1,41 @@
...
+    func keyStore(passphrase: String) async throws -> BIP32Keystore
+}
+
+actor EthBIP32Service: EthBIP32ServiceProtocol {
+    private var passphrase: String = ""
+    private var keystore: BIP32Keystore?
+    
+    private var erc20ApiService: EthApiServiceProtocol
+    
+    init(erc20ApiService: EthApiServiceProtocol) {
```
ChrisBenua: is naming correct? erc20ApiService and ethApiService are different entities

Adamant/Modules/Wallets/EthBIP32Service/EthBIP32Service.swift
```diff
@@ -0,0 +1,41 @@
...
+//  Copyright © 2025 Adamant. All rights reserved.
+//
+import Web3Core
+
+protocol EthBIP32ServiceProtocol {
+    func keyStore(passphrase: String) async throws -> BIP32Keystore
+}
+
+actor EthBIP32Service: EthBIP32ServiceProtocol {
+    private var passphrase: String = ""
```
ChrisBenua: I know that passphrase can be empty, but just for consistency with keystore property (if keystore is nil so should be passphrase and vice versa).
So maybe we should make it `String?` instead of representing empty state with empty string

Adamant/Modules/Wallets/EthBIP32Service/EthBIP32Service.swift
```diff
@@ -0,0 +1,41 @@
...
+    private var keystore: BIP32Keystore?
+    
+    private var ethApiService: EthApiServiceProtocol
+    
+    init(erc20ApiService: EthApiServiceProtocol) {
```
ChrisBenua: correct naming here too please

ChrisBenua (approved):

Lainaaa (approved):
Agree with Christian, otherwise lgtm.

Adamant/Modules/Wallets/EthBIP32Service/EthBIP32Service.swift
vovameister: good point, updated

PR #684, opened 2025-01-31
[trello.com/c/EB2cbIld] Bug: Coin selection in the wallet list

vovameister:
It seems the bug occurred due to an extra table update that came through the observer, sometimes observer have like 30 notification in like 0.5 seconds, and reloadData() perform 30 in this time. So now I add timer 1 sec for update to not let notification spam reloadData().

### DIFF
--- a/Adamant/Modules/Settings/VisibleWallets/VisibleWalletsViewController.swift
+++ b/Adamant/Modules/Settings/VisibleWallets/VisibleWalletsViewController.swift
@@ -79,7 +79,9 @@ final class VisibleWalletsViewController: KeyboardObservingViewController {
     required init?(coder aDecoder: NSCoder) {
         fatalError()
     }
-    
+    deinit {
+        NotificationCenter.default.removeObserver(self)
+    }
     override func viewDidLoad() {
         super.viewDidLoad()
         loadWallets()
--- a/Adamant/Modules/Settings/VisibleWallets/VisibleWalletsViewController.swift
+++ b/Adamant/Modules/Settings/VisibleWallets/VisibleWalletsViewController.swift
@@ -64,10 +64,11 @@ final class VisibleWalletsViewController: KeyboardObservingViewController {
     
     private let cellIdentifier = "cell"
     private let cellResetIdentifier = "cellReset"
+    private let taskManager = TaskManager()
     private var filteredWallets: [WalletCoreProtocol]?
     private var wallets: [WalletCoreProtocol] = []
     private var previousAppState: UIApplication.State?
-    
+    private var lastUpdateRequestTime: Date?
     // MARK: - Lifecycle
     
     init(visibleWalletsService: VisibleWalletsService, accountService: AccountService) {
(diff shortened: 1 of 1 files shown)


Adamant/Modules/Settings/VisibleWallets/VisibleWalletsViewController.swift
```diff
@@ -94,21 +94,6 @@ final class VisibleWalletsViewController: KeyboardObservingViewController {
...
-                queue: OperationQueue.main
-            ) { [weak self] _ in
-                MainActor.assumeIsolatedSafe {
-                    guard let self = self else { return }
-                    self.tableView.reloadData()
-                }
-            }
-        }
-        
         NotificationCenter.default.addObserver(
```
art-divin, line 125: I think it would be nice to add removal of `self` as an observer on deinit in this ViewController just as a good practice. What do you think @vovameister ?

```swift
deinit {
        NotificationCenter.default.removeObserver(self)
    }
```

vovameister: Good point. I added a change.

Adamant/Modules/Settings/VisibleWallets/VisibleWalletsViewController.swift
```diff
@@ -94,21 +96,6 @@ final class VisibleWalletsViewController: KeyboardObservingViewController {
...
-                forName: notification,
-                object: wallet,
-                queue: OperationQueue.main
-            ) { [weak self] _ in
-                MainActor.assumeIsolatedSafe {
-                    guard let self = self else { return }
-                    self.tableView.reloadData()
-                }
-            }
-        }
```
crypt-eng, line 110: Are balances updated without this observer? We display them on this screen, and we need to update the balances here

vovameister: what you think about add timeout (like 1 sec) for update from observer after didSelectRowAt?

crypt-eng: where exactly do you want to add the timeout?

crypt-eng: let's write code without timeouts

crypt-eng: If there is no fast fix, I think we can rewrite this screen to MVVM. There is no much of logics

ChrisBenua:
Are you sure you got the right ticket for the PR?
https://trello.com/c/Lw7fxven

Or am I missing out on something?

Adamant/Modules/Settings/VisibleWallets/VisibleWalletsViewController.swift
Lainaaa: I just tested, it does not update balances now.

Lainaaa (changes requested):

Adamant/Modules/Settings/VisibleWallets/VisibleWalletsViewController.swift
```diff
@@ -102,13 +105,23 @@ final class VisibleWalletsViewController: KeyboardObservingViewController {
...
-                    guard let self = self else { return }
-                    self.tableView.reloadData()
-                }
+                guard let self = self else { return }
+                
+                taskManager.clean()
+                lastUpdateRequestTime = Date()
+                
+                Task {
+                    try? await Task.sleep(nanoseconds: 1_000_000_000)
```
ChrisBenua, line 114: why not use debounce operator with Combine function for NotificationCenter? Way too complicated for simple logic

PR #685, opened 2025-01-31
[trello.com/c/dAj2Hypi] Architecture refactoring for secret wallets

Lainaaa:
Architecture refactoring.
Some wallet logic was moved to a new AccountWalletsViewModel to unload AccountViewController.
The new model has been created in a more appropriate format.
Replacing the notification center with void publisher in coin wallets to avoid singleton. 
AdamantVisibleWalletsService was changed to to cut down on the use of a particular type and not get hung up on it. Instead, the coin's string ID is now used.


Adamant/Modules/Account/AccountViewController/AccountWallets/AccountWalletsState.swift
```diff
@@ -0,0 +1,13 @@
...
+//  Adamant
+//
+//  Created by Dmitrij Meidus on 29.01.25.
+//  Copyright © 2025 Adamant. All rights reserved.
+//
+
+struct AccountWalletsState: Equatable {
+    var wallets: [WalletCollectionViewCell.Model]
+    
+    static let `default` = Self(wallets: [])
```
art-divin, line 12: Whenever I see ``` I think, can we name it differently so it won't collide with all the keywords of the language?
Maybe we could name it `shared` instead? As in "shared account wallets state".

For your consideration

Adamant/Modules/Account/AccountViewController/AccountWallets/AccountWalletsViewModel.swift
```diff
@@ -0,0 +1,74 @@
...
+        }
+    }
+
+    func updateInfo(for wallet: WalletService) {
+        if let index = state.wallets.firstIndex(where: { $0.coinID == wallet.core.tokenUnicID }) {
+            state.wallets[index].balance = wallet.core.wallet?.balance ?? 0
+            state.wallets[index].isBalanceInitialized = wallet.core.wallet?.isBalanceInitialized ?? false
+            state.wallets[index].notificationBadgeCount = wallet.core.wallet?.notifications ?? 0
+        } else {
+            let service = wallet.core
```
art-divin: You mind want to move line 50 here to line 45; so that in both branches `service` local variable would be used. Because to me, `wallet.core` naming is not the same as `service`, which that `core` actually is. Maybe even renaming that `service` to `coreService` would bring even more clarity?

Adamant/Modules/Account/Model.swift
```diff
@@ -0,0 +1,69 @@
+//
+//  Model.swift
```
art-divin: Filename can be more descriptive

Adamant/Modules/Account/AccountViewController/AccountViewController.swift
```diff
@@ -209,32 +202,17 @@ final class AccountViewController: FormViewController {
...
-                    MainActor.assumeIsolatedSafe {
-                        callback(notification)
-                    }
+        viewModel.$state
+            .removeDuplicates(by: { old, new in
+                old.wallets == new.wallets
+            })
+            .sink { [weak self] _ in
+                MainActor.assumeIsolatedSafe {
+                    guard let self = self else { return }
```
art-divin: Would subscribing to this $state publisher on the main queue eliminate the need in `MainActor.assumeIsolatedSafe`?

Adamant/Modules/Account/Model.swift
```diff
@@ -0,0 +1,69 @@
...
+        let index: Int
+        let coinID: String
+        let currencySymbol: String
+        let currencyImage: UIImage
+        let currencyNetwork: String
+        var isBalanceInitialized: Bool
+        var balance: Decimal?
+        var notificationBadgeCount: Int
+        
+        static let `default` = Model(
```
art-divin: I wonder, do we really "need" a default model, in a way, shall we ever use it, like, to present it to the user or anything like that? 
Or maybe there's a better way of organizing code around this model with an optional instead and associated values?

```swift
enum WalletCollectionViewModel {
    case empty
    case cell(model: WalletCollectionViewCell.Model)
}
```

What do you think @Lainaaa?

Adamant/Modules/Chat/ViewModel/ChatViewModel.swift
```diff
@@ -1467,7 +1467,7 @@ private extension ChatViewModel {
     
     func updateAttachmentButtonAvailability() {
         let isAnyWalletVisible = walletServiceCompose.getWallets()
-            .map { visibleWalletService.isInvisible($0.core) }
+            .map { visibleWalletService.isInvisible($0.core.tokenUnicID) }
```
art-divin, line 1470: Just wondering if `tokenUnicID` can be renamed to `tokenUniqueID` ?

Maybe out of the scope of this ticket, but it would make it clearer

Adamant/Modules/Settings/VisibleWallets/VisibleWalletsViewController.swift
```diff
@@ -161,8 +161,8 @@ final class VisibleWalletsViewController: KeyboardObservingViewController {
         }
     }
     
-    private func isInvisible(_ wallet: WalletCoreProtocol) -> Bool {
-        return visibleWalletsService.isInvisible(wallet)
+    private func isInvisible(_ walletTokenUnicID: String) -> Bool {
```
art-divin, line 164: `walletTokenUniqueID` ?

Adamant/Modules/Wallets/Adamant/AdmWalletService.swift
```diff
@@ -196,6 +203,9 @@ final class AdmWalletService: NSObject, WalletCoreProtocol, @unchecked Sendable
             object: self,
             userInfo: [AdamantUserInfoKey.WalletService.wallet: wallet]
         )
+        DispatchQueue.onMainThreadSyncSafe {
+            walletUpdateSender.send()
```
art-divin, line 207: subscribing on `mainQueue` would make it safer and remove assumption that the event was sent from non-main queue

Adamant/Modules/Wallets/Ethereum/EthWalletService.swift
```diff
@@ -150,6 +150,13 @@ final class EthWalletService: WalletCoreProtocol, @unchecked Sendable {
     let transactionFeeUpdated = Notification.Name("adamant.ethWallet.feeUpdated")
     let serviceStateChanged = Notification.Name("adamant.ethWallet.stateChanged")
     
+    @MainActor
+    private let walletUpdateSender = ObservableSender<Void>()
+    @MainActor
+    var walletUpdatePublisher: AnyObservable<Void> {
```
art-divin, line 156: since all of the wallest are relying on having this publisher, maybe it makes sense to make it a part of the `WalletCoreProtocol` ?

Adamant/Modules/Account/AccountViewController/AccountWallets/AccountWalletsState.swift
Lainaaa: I agree with you, and at first, it even confused me a bit. However, this is the existing code style, and I believe we should either stick to it or refactor everything consistently.

Adamant/Modules/Account/AccountViewController/AccountViewController.swift
Lainaaa: As I understand it [here](https://arc.net/l/quote/mdldginr), it uses that thread from which an event was sent. Even understanding that our class with publisher marked `@MainActor`, I prefer to explicitly specify using the main thread here and don't care in the future about it.

Adamant/Modules/Account/Model.swift
Lainaaa: I agree with you, and at first, it even confused me a bit. However, this is the existing code style, and I believe we should either stick to it or refactor everything consistently.

I would suggest putting this out for open discussion and deciding how we want to continue to organize this. I agree that this approach could be improved, I agree that this approach could be improved, but I also think it’s important to maintain consistency and avoid creating a mess.

Adamant/Modules/Chat/ViewModel/ChatViewModel.swift
Lainaaa: Agree

I could do it in this ticket if no one minds

Adamant/Modules/Wallets/Adamant/AdmWalletService.swift
Lainaaa: Could you clarify what exactly you mean.

Without `DispatchQueue.onMainThreadSyncSafe {` it shows me warning - 
`~/adamant-iOS/Adamant/Modules/Wallets/Adamant/AdmWalletService.swift:206:13 Main actor-isolated property 'walletUpdateSender' can not be referenced from a nonisolated context; this is an error in the Swift 6 language mode` 
to avoid it i decided to use it (not sure if it is the best approach, wdyt). Also it suggests to mark function as @MainActor but it starts to show me another warning, after that another error and etc, so i stopped with this solution

Adamant/Modules/Wallets/Ethereum/EthWalletService.swift
Lainaaa: It already has it:
WalletCoreProtocol.swift:297

Maybe I misunderstood yoy?

Adamant/Modules/Account/AccountViewController/AccountViewController.swift
Lainaaa: Also, I replaced
`MainActor.assumeIsolatedSafe {`
to
`.receive(on: DispatchQueue.main)`

Adamant/Modules/Wallets/Adamant/AdmWalletService.swift
PingPongEzZ: If u need to access something isolated from nonisolated, you could simply use ` Task { @MainActor(isolatedActor) in ... } `

Adamant/Modules/Account/AccountViewController/AccountWallets/AccountWalletsViewModel.swift
```diff
@@ -0,0 +1,74 @@
...
+    }
+
+    func updateInfo(for wallet: WalletService) {
+        let coreService = wallet.core
+        if let index = state.wallets.firstIndex(where: { $0.coinID == coreService.tokenUnicID }) {
+            state.wallets[index].balance = coreService.wallet?.balance ?? 0
+            state.wallets[index].isBalanceInitialized = coreService.wallet?.isBalanceInitialized ?? false
+            state.wallets[index].notificationBadgeCount = coreService.wallet?.notifications ?? 0
+        } else {
+
```
PingPongEzZ: Unneeded whitespace here. Just nit. 
Actually, swiftlint could have been marked this line as warning

Adamant/Modules/Wallets/Ethereum/EthWalletService.swift
art-divin: Nope, I missed this part, excuse me for the noise 🙏🏻

Adamant/Modules/Account/AccountViewController/AccountWallets/AccountWalletsViewModel.swift
art-divin: For the note, to ease things up for our code review process; we'd soon make warnings as errors in SwiftLint, add "autofix" in swiftlint and run the formatter as well on git hook. It would make code review much more pleasant

art-divin (approved):
Amazing work 🎆

vovameister (approved):

Adamant/Modules/Chat/ViewModel/ChatViewModel.swift
Lainaaa: https://github.com/Adamant-im/adamant-iOS/pull/690

PR #686, opened 2025-02-01
[trello.com/c/qGUKq5jZ] Sending ADM from chat: Do not display name for unnamed contacts

Lainaaa:
Added a check that the display of name+address only occurs if name and address are not equal. Since by default name is the address of adamant wallet, this check can work only when transferring in adamant network, as stated in the task.

### DIFF
--- a/Adamant/Modules/Wallets/TransferViewControllerBase.swift
+++ b/Adamant/Modules/Wallets/TransferViewControllerBase.swift
@@ -818,11 +818,10 @@ class TransferViewControllerBase: FormViewController {
             }
         }
         
-        let recipient: String
-        if let recipientName = recipientName {
+        var recipient: String = recipientAddress
+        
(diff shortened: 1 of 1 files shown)


Adamant/Modules/Wallets/TransferViewControllerBase.swift
```diff
@@ -818,11 +818,10 @@ class TransferViewControllerBase: FormViewController {
             }
         }
         
-        let recipient: String
-        if let recipientName = recipientName {
+        var recipient: String = recipientAddress
+        
+        if let recipientName = recipientName, recipientName != recipientAddress {
```
art-divin: ## Nit-pick

you could simply use `if let recipientName` here

ChrisBenua (approved):

art-divin (approved):

PR #687, opened 2025-02-01
[trello.com/c/WmyCJE1a] Settet donate as an non system wallet

PingPongEzZ:
I've changed .donate Contact to non system, because of some of this logic should be store in blockchain like it's renamed representation


art-divin (approved):
I've checked, seems like there are no side-effects which were relying on `donate` being a system contact.

Lainaaa (approved):

PR #689, opened 2025-02-03
[trello.com/c/1C0Wu6wt]: debug login flow speed up in CryptoSwift and LiskKit

ChrisBenua:
Replaced for-in range loops inside CryptoSwift and LiskKit. Changes inside CryptoSwift drastically improved BIP32/39 keystore creation for generating private/public keys before initialization.

Optimized `Scrypt.salsa20_8_typed(_:)`, `Scrypt.blockXor(_:_:_:)`, `HMAC.authenticate(_:)`, `PKCS5.PBKDF2.calculateBlock(_:blockNum:)`, `SHA2.process32(block:currentHash:)`, `SHA3.θ(_:)` and more

Also used much faster `UInt/Int(trunactingIfNeeded:)` than regular `UInt/Int()` constructor in DEBUG-builds.

Spend some time Profiling the app, I will open more PRs later.


art-divin (approved):
Well done and the difference is really visible 🤝

vovameister (approved):

PR #692, opened 2025-02-04
[trello.com/c/fzQGYRav]: fix disappearing cursor on amount field on macOS

ChrisBenua:
Idea 1: cursor is disappearing because UITextField shrink and becomes too narrow
Solution 1: adjust content hugging priority
Result 1: does not work

Idea 2: cursor is not actually disappearing, it just layouts incorrectly
Solution 2: adjust result `caretRect(for:)` in UITextField subclass
Result 2: does not work

Idea 3: combine both idea 1 and idea 2:
Result 3: works fine!

To do so I had to create almost a copy of `_FieldCell` class from Eureka, but with generic `T: UITextField` type along with `EdgeInsetTextField` that adjust caret size with given inset.

And make `SafeDecimalCell` a child of `CustomFieldCell<Double, EdgeInsetTextField>`.

https://github.com/user-attachments/assets/04f32259-5911-4ed1-bc06-a322a8cc1dd4

### DIFF
--- a/Adamant/Helpers/CustomFieldRow.swift
+++ b/Adamant/Helpers/CustomFieldRow.swift
@@ -1,0 +1,370 @@
+import Eureka
+import UIKit
+
+// Copy of _FieldCell from Eureka with generic descendant of UITextField class
+open class CustomFieldCell<T, TextFieldType: UITextField> : Cell<T>, UITextFieldDelegate, TextFieldCell where T: Equatable, T: InputTypeInitiable {
+
+    weak var _textField: TextFieldType!
+    public var textField: UITextField! {
+        _textField
+    }
+    weak var titleLabel: UILabel?
+
+    fileprivate var observingTitleText = false
+    private var awakeFromNibCalled = false
+
+    open var dynamicConstraints = [NSLayoutConstraint]()
+
+    private var calculatedTitlePercentage: CGFloat = 0.7
+
+    public required init(style: UITableViewCell.CellStyle, reuseIdentifier: String?) {
+
+        let textField = TextFieldType()
+        self._textField = textField
+        textField.translatesAutoresizingMaskIntoConstraints = false
+
+        super.init(style: style, reuseIdentifier: reuseIdentifier)
+
+        setupTitleLabel()
+
+        contentView.addSubview(titleLabel!)
+        contentView.addSubview(textField)
+
+        NotificationCenter.default.addObserver(forName: UIApplication.willResignActiveNotification, object: nil, queue: nil) { [weak self] _ in
+            guard let me = self else { return }
+            guard me.observingTitleText else { return }
+            me.titleLabel?.removeObserver(me, forKeyPath: "text")
+            me.observingTitleText = false
+        }
+        NotificationCenter.default.addObserver(forName: UIApplication.didBecomeActiveNotification, object: nil, queue: nil) { [weak self] _ in
+            guard let me = self else { return }
+            guard !me.observingTitleText else { return }
+            me.titleLabel?.addObserver(me, forKeyPath: "text", options: [.new, .old], context: nil)
+            me.observingTitleText = true
+        }
+
+        NotificationCenter.default.addObserver(forName: UIContentSizeCategory.didChangeNotification, object: nil, queue: nil) { [weak self] _ in
+            self?.setupTitleLabel()
+            self?.setNeedsUpdateConstraints()
+        }
+    }
+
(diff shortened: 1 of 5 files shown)


vovameister (approved):
Great job

Adamant/Helpers/CustomFieldRow.swift
```diff
@@ -0,0 +1,370 @@
...
+            textField.setContentHuggingPriority(priority, for: .horizontal)
+        }
+    }
+
+    open override func update() {
+        super.update()
+        detailTextLabel?.text = nil
+
+        if !awakeFromNibCalled {
+            if let title = row.title {
```
Lainaaa, line 97: let title = row.title ?? ""
            
                switch row.cellStyle {
                case .subtitle:
                    textField.textAlignment = .left
                    textField.clearButtonMode = .whileEditing
                default:
                    textField.textAlignment = title.isEmpty ? .left : .right
                    textField.clearButtonMode = title.isEmpty ? .whileEditing : .never
                }
                
Could we optimize a little the code lines in the way like above?

Adamant/Helpers/CustomFieldRow.swift
```diff
@@ -0,0 +1,370 @@
...
+                    textField.clearButtonMode = title.isEmpty ? .whileEditing : .never
+                }
+            } else {
+                textField.textAlignment = .left
+                textField.clearButtonMode = .whileEditing
+            }
+        } else {
+            textLabel?.text = nil
+            titleLabel?.text = row.title
+            if #available(iOS 13.0, *) {
```
Lainaaa, line 113: The minimal target is iOS 15 already

Adamant/Helpers/CustomFieldRow.swift
```diff
@@ -0,0 +1,370 @@
...
+                titleLabel.setContentHuggingPriority(
+                    UILayoutPriority(textField.contentHuggingPriority(for: .vertical).rawValue + 1), for: .vertical)
+                dynamicConstraints.append(NSLayoutConstraint(item: titleLabel, attribute: .centerX, relatedBy: .equal, toItem: textField, attribute: .centerX, multiplier: 1, constant: 0))
+            } else {
+                dynamicConstraints.append(NSLayoutConstraint(item: textField!, attribute: .centerY, relatedBy: .equal, toItem: contentView, attribute: .centerY, multiplier: 1, constant: 0))
+            }
+
+            if let imageView = imageView, let _ = imageView.image {
+                views["imageView"] = imageView
+                if let titleLabel = titleLabel, let text = titleLabel.text, !text.isEmpty {
```
Lainaaa, line 188: Could we use something like
```
                if let titleLabel = titleLabel, let text = titleLabel.text, !text.isEmpty {
                    dynamicConstraints += NSLayoutConstraint.constraints(withVisualFormat: "H:[imageView]-(15)-[titleLabel]-|", options: [], metrics: nil, views: views)
}
                    dynamicConstraints += NSLayoutConstraint.constraints(withVisualFormat: "H:[imageView]-(15)-[textField]-|", options: [], metrics: nil, views: views)
```

Lainaaa (commented):
This class is somewhat hard to read due to its complexity. There are several places where the code could be simplified, for example:

- Avoiding excessive use of if-else structures where possible.
- Improving readability by reducing nesting and making conditions clearer.

I've highlighted a few specific areas where refactoring might help. Let me know your thoughts!

Adamant/Helpers/CustomFieldRow.swift
ChrisBenua: I've copy pasted it 99% of it from Eureka (only thing I've done - make it generic over the subclasses of UITextField).

Are you sure we want to refactor Eureka's code right now?

Lainaaa: When I saw it was a copy from eureka I thought by the behavior, not a copy of the code :)

Lainaaa: I think we can keep it as it is. If possible, I’d ask to add a source link to the comment for reference.

Lainaaa (approved):

Adamant/Helpers/CustomFieldRow.swift
ChrisBenua: https://github.com/xmartlabs/Eureka/blob/028ef8e3191a256b8f6b8bb6b9496efcb0762dbc/Source/Rows/Common/FieldRow.swift#L132

PR #693, opened 2025-02-04
[trello.com/c/LC02sdnW]: optimize hexBytes calculation

ChrisBenua:
Before: 900+ ms on main thread, 1.12 total (release build)
After: 0 ms (even not shown in Xcode Instruments Time Profiler)

Problem 1:
```
 let indexStart = index(startIndex, offsetBy: i * 2) // O(n) complexity
 let indexEnd = index(indexStart, offsetBy: 2)
```

So the whole operation is O(n^2):

```
return (0..<count/2).reduce([]) { res, i in // O(n)
let indexStart = index(startIndex, offsetBy: i * 2) // O(n)
```

Moreover, then it requires to create substring:

```
let substring = self[indexStart..<indexEnd]
return res + [UInt8(substring, radix: 16) ?? 0]
```

So I simplified it using simple loop and `hexCharToUInt8` to convert each symbol to `UInt4` (but inside `UInt8`) and then join them in single `UInt8` using `bytes.append((highValue << 4) | lowValue)`

Before

After

### DIFF
--- a/CommonKit/Sources/CommonKit/Core/NativeAdamantCore.swift
+++ b/CommonKit/Sources/CommonKit/Core/NativeAdamantCore.swift
@@ -215,12 +215,22 @@ public final class NativeAdamantCore: AdamantCore {
 
 public extension String {
     func hexBytes() -> [UInt8] {
-        return (0..<count/2).reduce([]) { res, i in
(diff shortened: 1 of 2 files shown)


Lainaaa (approved):
Good job!

vovameister (approved):

PR #697, opened 2025-02-07
[trello.com/c/0dgGzsPE]: timeouts for sending messages

ChrisBenua:
I decided to add timeout to `HealthCheckWrapper` functions. As to me, adding timeout to `APICoreProtocol` makes no sense, because basic network timeout in Alamofire is much less - only 60 s. So the most actual waiting happens inside waiting for connection in `HealthCheckWrapper`.

Also, I added timeouts generating adamant-walled in CoinScript.rb.

As for now, there is the same timeout for message and message with files attachments. So `isFileTransfer` is how we decide whether is it simple message or message with files.

For handling timeouts I've added generic function `deadline` in CommonKit. It supports actor isolation as well.

Because not all services support timeouts, I've added new protocol `HealthCheckableTimeoutableError` and we add new function with timeout for `HealthCheckWrapper ` only when `Error: HealthCheckableTimeoutableError `


CommonKit/Sources/CommonKit/Helpers/Deadline.swift
```diff
@@ -0,0 +1,76 @@
...
+//  Deadline.swift
+//  CommonKit
+//
+//  Created by Christian Benua on 06.02.2025.
+//
+
+import Foundation
+import QuartzCore
+
+public func deadline<R>(
```
Lainaaa: Hey, this function is really well-written! Could you add a short comment explaining its purpose and usage? It would help the team understand its behavior. Thanks!

Lainaaa (approved):
Great job!

CommonKit/Sources/CommonKit/Helpers/Deadline.swift
PingPongEzZ: Let's not make a public func but put it in some king of AdamantUtilities

PingPongEzZ:
Maybe we can add timeout to all of the messages in the core of network layer?

CommonKit/Sources/CommonKit/Helpers/Deadline.swift
ChrisBenua: It's needed inside CommonKit, I can't place it in main target

ChrisBenua:
> Maybe we can add timeout to all of the messages in the core of network layer?

It should not be placed inside core network layer, because main "waiting" is inside `HealthCheckWrapper`. So real waiting should be in `HealthCheckWrapper`.

Also, in task were mentioned only messages (not for money transfers), so current logic is just fine, in my opinion

CommonKit/Sources/CommonKit/Helpers/Deadline.swift
ChrisBenua: Done

ChrisBenua: Made it internal

PingPongEzZ:
> So real waiting should be in `HealthCheckWrapper`.

Got it

PR #694, opened 2025-02-05
[trello.com/c/G77vaoVI]: trimming passphrase in all login scenarios

ChrisBenua:
Moved trimming logic to `loginWith(passphrase:)` function. Used only by `Go` button on keyboard and `Login` button.

PR #702, opened 2025-02-13
Fix of a same bug with trailing chars in case of "Return" button tapped on a keyboard

PingPongEzZ:
I've forgot about the other way of logging in from keyboard because of using the simulator only


ChrisBenua:
I've done the same PR already:
https://github.com/Adamant-im/adamant-iOS/pull/694

PR #703, opened 2025-02-13
[trello.com/c/pwKrXiiC] Update the logic of reading messages.

vovameister:
the current code is here: https://github.com/Adamant-im/adamant-iOS/pull/705
the following changes have been added:
1. All unnecessary functions responsible for and implementing full reading of the chatroom have been removed
2. Now chats that are manually marked as unread will not make the last chat transaction unread, this prevents potential bugs
3. Now messages in the chat view model have an unread parameter, which allows you to track which messages on the screen are unread
4. Relationships between transactions and reactions, between messages and reactions have been added to the database. This allows you to add unread reaction to message in the view model and work with the reading flow.
5. Now when you click on the alert notification about an incoming message, the chat opens and scrolls to this message

### DIFF
--- a/Adamant/Models/CoreData/MessageTransaction+CoreDataProperties.swift
+++ b/Adamant/Models/CoreData/MessageTransaction+CoreDataProperties.swift
@@ -19,6 +19,7 @@ extension MessageTransaction {
     @NSManaged public var isMarkdown: Bool
     @NSManaged public var message: String?
     @NSManaged public var reactionsData: Data?
+    @NSManaged public var richMessageTransactions: Set<RichMessageTransaction>?
     
     var reactions: Set<Reaction>? {
         get {
@@ -38,5 +39,8 @@ extension MessageTransaction {
             let data = try? PropertyListEncoder().encode(value)
             reactionsData = data
         }
+    }
+    func addToRichMessageTransactions(_ transaction: RichMessageTransaction) {
+        self.mutableSetValue(forKey: "richMessageTransactions").add(transaction)
     }
 }
--- a/Adamant/Modules/Chat/View/ChatViewController.swift
+++ b/Adamant/Modules/Chat/View/ChatViewController.swift
@@ -172,6 +172,16 @@ final class ChatViewController: MessagesViewController {
     
     override func viewWillDisappear(_ animated: Bool) {
         super.viewWillDisappear(animated)
+        let visibleIndexPaths = messagesCollectionView.indexPathsForVisibleItems
+        if viewAppeared {
+            for indexPath in visibleIndexPaths {
+                if let unreadSections = viewModel.unReadMesaggesIndexes,
+                   unreadSections.contains(indexPath.section) {
+                    
+                    viewModel.messageWasRead(index: indexPath.section)
+                }
+            }
+        }
         inputBar.isUserInteractionEnabled = false
         inputBar.inputTextView.resignFirstResponder()
     }
@@ -200,7 +210,6 @@ final class ChatViewController: MessagesViewController {
             cell.isSelected = false
             viewModel.needToAnimateCellIndex = nil
         }
-        
         super.collectionView(collectionView, willDisplay: cell, forItemAt: indexPath)
     }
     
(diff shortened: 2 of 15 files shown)


Adamant/Models/CoreData/MessageTransaction+CoreDataProperties.swift
```diff
@@ -39,4 +40,7 @@ extension MessageTransaction {
             reactionsData = data
         }
     }
+    func addToRichMessageTransactions(_ transaction: RichMessageTransaction) {
+        self.mutableSetValue(forKey: "richMessageTransactions").add(transaction)
```
ChrisBenua, line 44: Oh, I see why you had to do like this. If you mutate richMessageTransactions, nothing will be saved to database.

The same thing would happen if you've used raw NSMutableSet

Comment is for other reviewers)

Adamant/Modules/Chat/View/ChatViewController.swift
```diff
@@ -215,13 +224,22 @@ final class ChatViewController: MessagesViewController {
     
     override func scrollViewDidScroll(_ scrollView: UIScrollView) {
         super.scrollViewDidScroll(scrollView)
+        let visibleIndexPaths = messagesCollectionView.indexPathsForVisibleItems
```
ChrisBenua, line 227: same code is repeated twice. Please extract it to function

Adamant/Modules/ChatsList/ChatListViewController.swift
```diff
@@ -70,6 +70,11 @@ final class ChatListViewController: KeyboardObservingViewController {
     var searchController: UISearchController?
     
     private var transactionsRequiringBalanceUpdate: [String] = []
+    private lazy var chatsManuallyMarkedAsUnread: Set<Int> = Set() {
```
ChrisBenua, line 73: do we really need lazy var here?

Adamant/Services/DataProviders/InMemoryCoreDataStack.swift
```diff
@@ -20,6 +20,8 @@ final class InMemoryCoreDataStack: CoreDataStack {
         
         let description = NSPersistentStoreDescription()
         description.type = NSInMemoryStoreType
+        description.shouldMigrateStoreAutomatically = true
```
ChrisBenua, line 23: Why? I don't think that NSInMemoryStoreType requires migrations at all

Adamant/Modules/ChatsList/ChatListViewController.swift
vovameister: yes because NSFetchedResultsController(setBadgeValue(unreadController?.fetchedObjects?.count)) will be available later

ChrisBenua: I don't get it. You set lazy var to empty set, for what?

Adamant/Services/DataProviders/InMemoryCoreDataStack.swift
vovameister: fixed

Adamant/Modules/ChatsList/ChatListViewController.swift
vovameister: oh sorry I might use another value before, and it crushed the app on init. I will remove lazy

Adamant/Services/DataProviders/InMemoryCoreDataStack.swift
Lainaaa: @vovameister take a look, please!

Adamant/Modules/Chat/ViewModel/ChatMessageFactory.swift
```diff
@@ -541,6 +542,21 @@ private extension ChatMessageFactory {
             ]
         ))
     }
+    func checkTransactionForUnreadReaction(transaction: ChatTransaction) -> Bool {
+        if let messageTransaction = transaction as? MessageTransaction,
```
Lainaaa, line 546: Maybe as part of this тask we can take richMessageTransactions out to ChatTransaction and avoid code duplication?

Adamant/Modules/Chat/ViewModel/ChatViewModel.swift
```diff
@@ -400,18 +401,15 @@ final class ChatViewModel: NSObject {
         guard let address = chatroom?.partner?.address else { return }
         chatsProvider.setChatPositon(for: address, position: offset.map { Double.init($0) })
     }
-    
-    func entireChatWasRead() {
+    func messageWasRead(index: Int) {
```
Lainaaa, line 404: it sounds like it should return a bool - was a message read or not
Could we call it "markMessageAsRead" instead?

Lainaaa (approved):
lgtm
However, check, please, Christian comments

PR #706, opened 2025-02-18
[trello.com/c/NOMM1JYt] fix Lags on Chats screen

vovameister:
Lags on Chats screen when message's side menu opened and chats are updating
This bug occurred because we separately reload all visible cells in the method refreshDatesIfNeeded(), now we will not reload swiped cell in this method, but we will reload this cell after the end of the swipe

### DIFF
--- a/Adamant/Modules/ChatsList/ChatListViewController.swift
+++ b/Adamant/Modules/ChatsList/ChatListViewController.swift
@@ -130,7 +130,7 @@ final class ChatListViewController: KeyboardObservingViewController {
     
     private var loadNewChatTask: Task<(), Never>?
     private var subscriptions = Set<AnyCancellable>()
-    
+    private var swipedIndex: IndexPath?
     // MARK: Init
     
     init(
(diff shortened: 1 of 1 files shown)


Adamant/Modules/ChatsList/ChatListViewController.swift
```diff
@@ -360,12 +360,13 @@ final class ChatListViewController: KeyboardObservingViewController {
...
     private func refreshDatesIfNeeded() {
         guard !isBusy,
-              let indexPaths = tableView.indexPathsForVisibleRows
+              var indexPaths = tableView.indexPathsForVisibleRows
         else {
             return
         }
         
         lastDatesUpdate = Date()
+        indexPaths.removeAll { $0 == swipedIndex }
```
Lainaaa, line 369: What do you think to invoke ```.removeAll { $0 == swipedIndex }``` where you init a variable and instead of var use let?

Lainaaa (approved):
lgtm

Adamant/Modules/ChatsList/ChatListViewController.swift
```diff
@@ -1106,7 +1109,12 @@ extension ChatListViewController {
         
         return UISwipeActionsConfiguration(actions: actions)
     }
-    
+    func tableView(_ tableView: UITableView, didEndEditingRowAt indexPath: IndexPath?) {
+        swipedIndex = nil
```
ChrisBenua, line 1113: does this method get called on regular `reloadData` call? Or do we need to reset `swipedIndex` before each `reloadData` call?

ChrisBenua (approved):
lgtm

PR #708, opened 2025-02-21
[trello.com/c/7YVA2NTe]: cancel file uploading

ChrisBenua:
Cancelling file upload.

I've introduced new `ChatAction` - `cancelUploading(messageId: String, file: ChatFile)`.

For handling - `ChatFileService.cancelUpload`. To handle it in `ChatFileService` I've introduced  `var uploadTasks: [String: Task<UploadFileResult, Error>]`.

For removing single file from message - `removeFromRichFile` function was introduced.

Moved DB-operations for hiding message from `func hideMessage(id: String)` in `ChatViewModel` to `func removeMessage(with id: String)` in `AdamantChatProvider`.

For now there is no specific UI. Just tapping on uploading file - cancels uploading and removes file.

### DIFF
--- a/Adamant/Modules/Chat/ViewModel/ChatFileService.swift
+++ b/Adamant/Modules/Chat/ViewModel/ChatFileService.swift
@@ -24,11 +24,13 @@ private struct FileMessage {
     var files: [FileUpload]
     var message: String?
     var txId: String?
+    var adamantMessage: AdamantMessage?
 }
 
 @MainActor
 final class ChatFileService: ChatFileProtocol, Sendable {
     typealias UploadResult = (decodedData: Data, encodedData: Data, nonce: String, cid: String)
+    typealias UploadFileResult = (file: UploadResult, preview: UploadResult?)
     
     // MARK: Dependencies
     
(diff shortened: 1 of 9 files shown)


Adamant/Modules/Chat/ViewModel/ChatFileService.swift
```diff
@@ -1136,12 +1214,35 @@ private extension ChatFileService {
             throw FileManagerError.cantEncryptFile
         }
         
-        let cid = try await filesNetworkManager.uploadFiles(
+        try Task.checkCancellation()
+        
+        let result = await filesNetworkManager.uploadFiles(
```
art-divin, line 1215: `ChatFileService` is marked with `@MainActor` and so this await blocks the main thread until the file is uploaded (supposedly). Could you please verify that this does not happen and if so, add `Task` to wrap this async call. Also considering this is a service, does it really make sense to run it as the main actor?

art-divin: Thanks to @ChrisBenua, I have educated myself using [this video](https://www.youtube.com/watch?v=H_K-us4-K7s). No issue here ✅

Adamant/Modules/Chat/ViewModel/ChatFileService.swift
```diff
@@ -916,34 +924,78 @@ private extension ChatFileService {
...
+    func processFilesUpload(
+        fileMessage: inout FileMessage,
+        chatroom: Chatroom?,
+        keyPair: Keypair,
+        storageProtocol: NetworkFileProtocolType,
+        ownerId: String,
+        partnerAddress: String,
+        saveEncrypted: Bool,
+        txId: String,
+        richFiles: inout [RichMessageFile.File],
```
Lainaaa: I don't see if richFiles is used anywhere in this function.

Lainaaa (approved):
Should be good.

PingPongEzZ (approved):

PR #710, opened 2025-02-24
[trello.com/c/QCnKo03A] Adding the logic of timestampMs

PingPongEzZ:
I've added a new field for Transaction DTO which are mapped in ChatTransaction entities. Now we have only timestamp field, but we need to have backward compatibility so I setting now the timestamp value from the backend multiplied by 1000 if there is no timestampMs in this DTO. 
And I've changed comparing descriptor of the CoreData's ChatTransactions to make comparing more precisely because we are comparing the timestampMs now

### DIFF
--- a/Adamant/Services/DataProviders/AdamantTransfersProvider.swift
+++ b/Adamant/Services/DataProviders/AdamantTransfersProvider.swift
@@ -443,7 +443,7 @@ extension AdamantTransfersProvider {
         let transaction = TransferTransaction(context: context)
         transaction.amount = amount as NSDecimalNumber
         transaction.date = Date() as NSDate
-        transaction.timestampMs = transaction.date?.timeIntervalMillisecondsSince1970 ?? .zero
+        transaction.timestampMs = transaction.timeIntervalMillisecondsSince1970
         transaction.recipientId = recipient
         transaction.senderId = loggedAccount.address
         transaction.type = Int16(TransactionType.chatMessage.rawValue)
@@ -637,7 +637,7 @@ extension AdamantTransfersProvider {
         let transaction = TransferTransaction(context: context)
         transaction.amount = amount as NSDecimalNumber
         transaction.date = Date() as NSDate
-        transaction.timestampMs = transaction.date?.timeIntervalMillisecondsSince1970 ?? .zero
+        transaction.timestampMs = transaction.timeIntervalMillisecondsSince1970
         transaction.recipientId = recipient
         transaction.senderId = loggedAccount.address
         transaction.type = Int16(TransactionType.send.rawValue)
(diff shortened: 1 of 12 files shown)


Adamant/Services/DataProviders/AdamantTransfersProvider.swift
```diff
@@ -446,6 +443,7 @@ extension AdamantTransfersProvider {
         let transaction = TransferTransaction(context: context)
         transaction.amount = amount as NSDecimalNumber
         transaction.date = Date() as NSDate
+        transaction.timestampMs = transaction.date?.timeIntervalMillisecondsSince1970 ?? .zero
```
ChrisBenua: should we extract `transaction.date?.timeIntervalMillisecondsSince1970 ?? .zero` to extension?

CommonKit/Sources/CommonKit/Helpers/Date+TimestampMs.swift
```diff
@@ -0,0 +1,20 @@
...
+import Foundation
+
+public extension NSDate {
+    var timeIntervalMillisecondsSince1970: Int64 {
+        let formatter = DateFormatter()
+        formatter.dateFormat = "SSS"
+        guard let milliseconds = Int64(formatter.string(from: self as Date)) else {
+            return Int64(timeIntervalSince1970) * 1000
+        }
+        return Int64(timeIntervalSince1970) * 1000 + milliseconds
```
ChrisBenua: `self.timeIntervalSince1970 * 1000.0`

there is no reason to use DateFormatters here

PingPongEzZ: We need to get a millisecond value from the instance of the NSDate. There is no property in Date that represents the millisecond value. But if you have a suggestion it would be great to see it.
The main reason of a dateformatter here is that we need to know exact milliseconds like `435`, but not `000` as default for each date instance.
We need to compare a few different transactions sent in the range of one-second like:

1st sent at 28.02.25 15:43:44.210
2d sent at 28.02.25 15:43:44.765
3d sent at 28.02.25 15:43:44.998

PingPongEzZ: That's just a crutch which I've found for this idea

Adamant/Services/DataProviders/AdamantTransfersProvider.swift
PingPongEzZ: This is already and extension for a Date class - timeIntervalMillisecondsSince1970.
I didn't want to add some new extensions for out. But yeah, could be good to add it into a CoinTransaction entity

CommonKit/Sources/CommonKit/Helpers/Date+TimestampMs.swift
PingPongEzZ: Didn't know about that thing with the milliseconds in the timeIntervalSince1970. Thx to @ChrisBenua

vovameister (approved):

Lainaaa (commented):
lgtm

Lainaaa (approved):

PR #714, opened 2025-02-28
[trello.com/c/lwOnjxPQ] Sending files indicator 🕝 in the chat list + ✏️

vovameister:
Added display of a message in the chatlist even if it is still sending or a if its a draft, 🕝 is displayed before the sent message and ✏️ before the draft. Now sent messages are saved in the database earlier so that we can display them as the latest transaction.
Снимок экрана 2025-02-27 в 23 30 40
Снимок экрана 2025-02-28 в 15 51 04


PingPongEzZ (approved):

Adamant/Modules/ChatsList/ChatListViewController.swift
```diff
@@ -1034,7 +1045,53 @@ extension ChatListViewController {
...
+
+        if chatPreservation.getReplyMessage(address: address, thenRemoveIt: false) != nil {
+                let replyImageAttachment = NSTextAttachment()
+                replyImageAttachment.image = UIImage(systemName: "arrowshape.turn.up.left")?.withTintColor(.adamant.primary)
+                replyImageAttachment.bounds = CGRect(x: .zero, y: -3, width: 23, height: 20)
+
+                descriptionParts.append(NSAttributedString(attachment: replyImageAttachment))
+            }
+        if let files = chatPreservation.getPreservedFiles(for: address, thenRemoveIt: false), !files.isEmpty {
+                let mediaCount = files.filter { $0.type.isMedia }.count
```
ChrisBenua: May be it will be better to use https://developer.apple.com/documentation/swift/sequence/count(where:) here?

Adamant/Modules/ChatsList/ChatListViewController.swift
```diff
@@ -1034,7 +1045,53 @@ extension ChatListViewController {
...
+                }
+            }
+
+        if let preservedMessage = chatPreservation.getPreservedMessageFor(address: address, thenRemoveIt: false) {
+            let processedMessage = MessageProcessHelper.process(preservedMessage)
+            descriptionParts.append(NSAttributedString(string: processedMessage))
+        }
+        let hasTextContent = descriptionParts.contains { !$0.string.trimmingCharacters(in: .whitespacesAndNewlines).isEmpty }
     
+        guard hasTextContent else { return nil }
```
ChrisBenua: inline guard case?

`guard descriptionParts.contains { !$0.string.trimmingCharacters(in: .whitespacesAndNewlines).isEmpty }`

Adamant/Modules/ChatsList/ChatListViewController.swift
vovameister: super minor update but as you wish;)

Adamant/Modules/ChatsList/ChatListViewController.swift
vovameister: got it

ChrisBenua (approved):

PR #717, opened 2025-03-04
[trello.com/c/GpoP9FLk] Added logic of cancelling of the current tasks

PingPongEzZ:
This PR is about refactoring and optimization. 

What has been done: 
1. Mainly is that I've added logic to cancel current tasks of current user. It led to some data race errors before because of following logic:
    1. User have triggered initial chats update.
    2. User logged out before request triggered by the 1st step haven't been finished yet.
    3. CoreDate have updated.
    4. Request triggered from 1st step updates  CoreData. 
    5. User logs in to the other wallet
    6. As result old messages from old core data of the previous wallet
2. I've made CoreData invalidation due to logging out event to be synchronious 
3. I've made updating of the `isInitiallySynced` value on user logging in event to pretend every new logging-in as the first logging-in one lifecycle of the app.
4. Introduced `AdamantApiTask` to manage and indentify the lifecycle of the Tasks created in `AdamantApiService`.
`AdamantApiTask` uses `id` to identify each tasks, for store and remove it from the tasks storage in `AdamantApiService`. It also conforms new custom `CancellableTask` interface to avoid Generic restrictions of `request<Output>` function

### DIFF
--- a/Adamant/ServiceProtocols/DataProviders/CoreDataStack.swift
+++ b/Adamant/ServiceProtocols/DataProviders/CoreDataStack.swift
@@ -11,4 +11,5 @@ import CoreData
 
 protocol CoreDataStack: Sendable {
     var container: NSPersistentContainer { get }
+    func clearCoreData()
 }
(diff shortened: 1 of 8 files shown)


Adamant/ServiceProtocols/DataProviders/CoreDataStack.swift
vovameister, line 1: I would move to protocol CoreDataStack in InMemoryCoreDataStack on top to not create extra files especially for such small protocols

vovameister (approved):

Adamant/ServiceProtocols/DataProviders/CoreDataStack.swift
art-divin: I would leave this one, separate files get really useful when project is moved to a modular architecture and encapsulation is enforced

Lainaaa (approved):
lgtm

PR #718, opened 2025-03-05
[trello.com/c/ZWei9TCF] fix visual bug with pendings

vovameister:
The initial status for transactions is .notInitiated. After the first response from the server, the status is assigned to .registered. For erc20 transactions, I changed nil scenario, since they are rich transactions and do not receive intermediate responses from the server.
Adamant/Modules/Wallets/TransactionTableViewCell.swift this cell should not show clock, so i change it


PingPongEzZ (approved):

PingPongEzZ (approved):

art-divin (approved):
Nice work!

PR #719, opened 2025-03-07
[trello.com/c/F7EV0pKq] Rework of the popup

PingPongEzZ:
I've updated a few things with the "No active nodes" popup.
Now it'd be presented after pushing the chat screen to inform user why new messages cannot be sent and updated from the chain

### DIFF
--- a/Adamant/Modules/Chat/View/Managers/ChatDialogManager.swift
+++ b/Adamant/Modules/Chat/View/Managers/ChatDialogManager.swift
@@ -192,7 +192,7 @@ private extension ChatDialogManager {
     
     func showNoActiveNodesAlert() {
         let alert = UIAlertController(
-            title: "",
+            title: .adamant.chat.noActiveNodesTitle,
             message: .adamant.chat.noActiveNodes,
             preferredStyleSafe: .alert,
             source: nil
@@ -201,13 +201,15 @@ private extension ChatDialogManager {
         alert.addAction(
             .init(
                 title: .adamant.chat.reviewNodesList,
-                style: .default,
+                style: .destructive,
                 handler: { [weak self] _ in
                     self?.viewModel.didTapAdmNodesList.send(())
                 }
             )
         )
-        alert.addAction(.init(title: .adamant.alert.cancel, style: .cancel))
+        let cancelButton = UIAlertAction(title: .adamant.alert.cancel, style: .default)
+        cancelButton.setValue(UIColor.blue, forKey: "titleTextColor")
+        alert.addAction(cancelButton)
         alert.modalPresentationStyle = .overFullScreen
         dialogService.present(alert, animated: true, completion: nil)
     }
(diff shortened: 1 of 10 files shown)


Adamant/Modules/Chat/View/Managers/ChatDialogManager.swift
```diff
@@ -201,13 +201,15 @@ private extension ChatDialogManager {
...
-                style: .default,
+                style: .destructive,
                 handler: { [weak self] _ in
                     self?.viewModel.didTapAdmNodesList.send(())
                 }
             )
         )
-        alert.addAction(.init(title: .adamant.alert.cancel, style: .cancel))
+        let cancelButton = UIAlertAction(title: .adamant.alert.cancel, style: .default)
+        cancelButton.setValue(UIColor.blue, forKey: "titleTextColor")
```
art-divin: Hey;

Really unsure about setting title text color this way. It does not look as good as when using default tint colour; also it does not change the title text colour on macOS version for this button.

Adamant/Modules/Chat/View/Managers/ChatDialogManager.swift
```diff
@@ -201,13 +201,15 @@ private extension ChatDialogManager {
         alert.addAction(
             .init(
                 title: .adamant.chat.reviewNodesList,
-                style: .default,
+                style: .destructive,
```
art-divin, line 204: Destructive action is not what actually happens, when we tell the user to review the active node list. User is not destroying anything, why not to stick to the default action instead?

Adamant/Modules/Chat/View/Managers/ChatDialogManager.swift
PingPongEzZ: I'm not sure about macos, forgot about it, but this is only way to change the color of the text for actions in the alert. I'll check it for mac

Adamant/Modules/Chat/View/Managers/ChatDialogManager.swift
PingPongEzZ: Nope. Alexei said we need to make this button red, that's why i use it here. To not change the color as i did it for blue cancel button

Adamant/Modules/Chat/View/Managers/ChatDialogManager.swift
PingPongEzZ: Just a business requirements

art-divin (approved):

Adamant/Modules/Chat/View/Managers/ChatDialogManager.swift
art-divin: As discussed we are good to not set colour for now, rather investigate how to do it properly using UIAppearance proxy instead.

Lainaaa (approved):
lgtm

PR #720, opened 2025-03-11
[trello.com/c/AvS5QpYb] Updating again the popup logic

PingPongEzZ:
Fixing a bug with broken selected wallet after few time of being idle. The bug was about updating the wallet controller after some time. But after the updates of this section index has been invalidating. I've found this and made an index update after the data reloading. Also I've invistigated that someone has broke the currencySymbol logic, which where made to compare current selected wallet with node availability


Lainaaa (approved):
lgtm

art-divin (approved):

PR #732, opened 2025-03-14
[trello.com/c/esjdkfhK] Bug/UI: Extra scroll on many new messages

vovameister:
Сlean and effective bug fix without unnecessary changes, as in the first version

### DIFF
--- a/Adamant/Modules/Chat/View/Subviews/ChatMessagesCollection.swift
+++ b/Adamant/Modules/Chat/View/Subviews/ChatMessagesCollection.swift
@@ -43,14 +43,20 @@ final class ChatMessagesCollectionView: MessagesCollectionView {
         }
     }
     
-    func reloadData(newIds: [String]) {
-        guard newIds.last == currentIds.last || newIds.first != currentIds.first else {
+    func reloadData(newIds: [String], isOnBottom: Bool) {
+        let hasNewMessagesAtTop = newIds.first != currentIds.first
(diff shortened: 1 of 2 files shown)


Adamant/Modules/Chat/View/Subviews/ChatMessagesCollection.swift
```diff
@@ -43,14 +43,20 @@ final class ChatMessagesCollectionView: MessagesCollectionView {
         }
     }
     
-    func reloadData(newIds: [String]) {
-        guard newIds.last == currentIds.last || newIds.first != currentIds.first else {
+    func reloadData(newIds: [String], isOnBottom: Bool) {
```
Lainaaa, line 46: The isOnBottom parameter is too specific. Consider renaming it to something more abstract

Lainaaa (dismissed):

Adamant/Modules/Chat/View/Subviews/ChatMessagesCollection.swift
vovameister: Look inside the function you have the following logical chain: if new messages appeared at the top, we add an offset of 100%, if at the bottom, then we look at the bottom of the chat whether we are there or not, if not, then we do not update.
I do not see the point in these logical reasonings to bring the latter to a more abstract level
This only adds confusion

PingPongEzZ (approved):

art-divin (approved):

PR #733, opened 2025-03-17
[trello.com/c/hzqPFC5l] Update "Send Crypto" screen to use with secret wallets

Lainaaa:
You can find description of the main part in comments.


Adamant/Modules/Wallets/TransferViewControllerBase.swift
```diff
@@ -1023,6 +1037,28 @@ class TransferViewControllerBase: FormViewController {
...
+            }
+        case .type:
+            return LabelRow {
+                $0.title = BaseRows.type.localized
+                $0.tag = BaseRows.type.tag
+                $0.disabled = true
+                
+                $0.value = secretWalletViewModel.state.currentWallet?.name ?? String.localized("SecretWallets.Menu.Regular", comment: "Regular Wallet")
+                
+                for wallet in secretWalletManager.getRegularWallet().sorted(includeInvisible: false) where wallet.core.wallet?.address == walletCore.wallet?.address {
```
Lainaaa, line 1056: This view controller is used on two different screens. On the account tab, you can access it by tapping "Send (cryptoTokenName)". In an open chat, you can find it by tapping the plus button and choosing "Transfer Crypto".

Even though the same component is used in both places, the context for secret wallets differs: in the chat, only the regular wallet can be used, whereas on the account tab, it doesn’t matter if the wallet is secret or regular.

My solution is simple:
By default, we show the name from the current wallet (which might be secret). However, if the regular wallet has the same address as the current wallet in the view controller, we change the display name to "Regular Wallet".

vovameister (approved):
well done

PR #736, opened 2025-03-17
[trello.com/c/jKqpxiCa] Improve in-chat scroll position | "New messages" separator

vovameister:
add separate line for new messages, used Message collection view footer for it 
Снимок экрана 2025-03-05 в 15 26 22 2

### DIFF
--- a/Adamant/Modules/Chat/View/Managers/ChatDisplayManager.swift
+++ b/Adamant/Modules/Chat/View/Managers/ChatDisplayManager.swift
@@ -67,6 +67,23 @@ final class ChatDisplayManager: MessagesDisplayDelegate {
         }
     }
     
+    nonisolated func messageFooterView(
+        for indexPath: IndexPath,
+        in messagesCollectionView: MessagesCollectionView
+    ) -> MessageReusableView {
+        DispatchQueue.onMainThreadSyncSafe {
+            guard let separatorIndex = viewModel.separatorIndex, indexPath.section == separatorIndex else {
+                return MessageReusableView()
+            }
+            
+            let footer = messagesCollectionView.dequeueReusableFooterView(
+                NewMessagesCell.self,
+                for: indexPath
+            )
+            return footer
+        }
+    }
+    
     nonisolated func enabledDetectors(
         for _: MessageType,
         at _: IndexPath,
--- a/Adamant/Modules/Chat/ViewModel/ChatViewModel.swift
+++ b/Adamant/Modules/Chat/ViewModel/ChatViewModel.swift
@@ -54,7 +54,7 @@ final class ChatViewModel: NSObject {
     private var isLoading = false
     var separatorIndex: Int?
     var separatorId: String?
-    var didAddSeparator: Bool = false
+    var isSeparatorAdded: Bool = false
     
     private var isNeedToLoadMoreMessages: Bool {
         get async {
--- a/Adamant/Modules/Chat/ViewModel/ChatViewModel.swift
+++ b/Adamant/Modules/Chat/ViewModel/ChatViewModel.swift
@@ -1104,8 +1104,8 @@ private extension ChatViewModel {
             .store(in: &subscriptions)
         $messages
             .removeDuplicates()
-            .sink { _ in
-                self.updateSeparatorIndex()
+            .sink { [weak self] _ in
+                self?.updateSeparatorIndex()
             }
             .store(in: &subscriptions)
         chatFileService.updateFileFields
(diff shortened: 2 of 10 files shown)


Adamant/Modules/Chat/View/Managers/ChatDisplayManager.swift
Lainaaa, line 1: Why do we use @MainActor here while every method is nonisolated?

Adamant/Modules/Chat/ViewModel/ChatViewModel.swift
```diff
@@ -1703,6 +1713,31 @@ private extension ChatViewModel {
             shouldScrollToBottom = true
         }
     }
+    
+    func updateSeparatorId() {
```
Lainaaa, line 1717:
```suggestion
    private func updateSeparatorId() {
```

Adamant/Modules/Chat/ViewModel/ChatViewModel.swift
```diff
@@ -1703,6 +1713,31 @@ private extension ChatViewModel {
...
+            separatorIndex = nil
+            return
+        }
+        
+        didAddSeparator = true
+        separatorId = firstUnreadId
+        updateSeparatorIndex()
+    }
+
+    func updateSeparatorIndex() {
```
Lainaaa, line 1731:
```suggestion
    private func updateSeparatorIndex() {
```

Adamant/Modules/Chat/ViewModel/ChatViewModel.swift
```diff
@@ -52,6 +52,9 @@ final class ChatViewModel: NSObject {
     private var timerSubscription: AnyCancellable?
     private var messageIdToShow: String?
     private var isLoading = false
+    var separatorIndex: Int?
+    var separatorId: String?
+    var didAddSeparator: Bool = false
```
Lainaaa, line 57:
```suggestion
    var isSeparatorAdded: Bool = false
```

Adamant/Modules/Chat/ViewModel/ChatViewModel.swift
Lainaaa, line 1717: There's some very complicated logic here. Can we simplify it somehow? What if we create a custom structure instead of three variables? Why do we need two methods especially if one method calls the other? Why do we need to call these methods separately in subscriptions if the subscription is the same, only the processing is different?

Lainaaa (changes requested):

Adamant/Modules/Chat/View/Managers/ChatDisplayManager.swift
vovameister: idk, I dont like to refactor this class, I we will remove message Kit some day it will be remove too

Adamant/Modules/Chat/ViewModel/ChatViewModel.swift
vovameister: its private extension

Adamant/Modules/Chat/ViewModel/ChatViewModel.swift
vovameister: its private extension

Adamant/Modules/Chat/ViewModel/ChatViewModel.swift
vovameister: I wouldn't make a separate structure for 3 variables, one of which will change once. 1st method should be called only once, then it will always call return after guard, and the second method is needed to update the separator index, it should be called constantly

PingPongEzZ (approved):

Lainaaa (changes requested):
The project fails to build.

Screenshot 2025-03-19 at 10 53 44 AM

Adamant/Modules/Chat/ViewModel/ChatViewModel.swift
```diff
@@ -52,6 +52,9 @@ final class ChatViewModel: NSObject {
     private var timerSubscription: AnyCancellable?
     private var messageIdToShow: String?
     private var isLoading = false
+    var separatorIndex: Int?
+    var separatorId: String?
+    var isSeparatorAdded: Bool = false
```
Lainaaa: ViewModel should not manage UI logic like was separator added or not

Lainaaa (approved):
I am not confident about code, it seems that might be better approach, but i am not able to find it!

 but it 100% works! 
 
@art-divin I would appreciate for your review here!

PR #739, opened 2025-03-18
[trello.com/c/9vzwpqen] Fixed Issue with File Upload

art-divin:
## Context

Due to refactoring of how uploadable files are stored in a dictionary rather than passed as inout arguments,
and the fact that value types are used,
it is necessary to update stored values after processing.

## Video

|Video|
:-:
Simulator Screen Recording - iPhone 16 Pro - 2025-03-19 at 03 17 20


vovameister (approved):

Lainaaa (approved):
lgtm

PR #741, opened 2025-03-19
[trello.com/c/Aw0Vfnmd] {Secret wallets} UI: Show Transactions 🔐 1️⃣

Lainaaa:


Adamant/Modules/Wallets/TransactionsListViewControllerBase.swift
```diff
@@ -114,7 +117,7 @@ class TransactionsListViewControllerBase: UIViewController {
         super.viewDidLoad()
 
         navigationItem.largeTitleDisplayMode = .never
-        navigationItem.title = String.adamant.transactionList.title
+        navigationItem.title = String.adamant.transactionList.title + (secretWalletsViewModel.state.currentWallet?.name ?? "")
```
art-divin, line 120: Nit: you might want to consider to have an extension on `secretWalletsViewModel.state` which would return `currentWalletName` to simplify code here

Adamant/Services/AdamantCoinStorageService.swift
```diff
@@ -104,24 +107,49 @@ private extension AdamantCoinStorageService {
         )
         .sink { [weak self] notification in
             let changes = notification.managedObjectContextChanges(of: CoinTransaction.self)
-
+            
+            guard self != nil, self?.coinId != nil && self?.coinAddress != nil else {
```
art-divin: guard let self else { return } would eliminate the need to force-uwrap self later

art-divin (approved):

Adamant/Modules/Wallets/TransactionsListViewControllerBase.swift
Lainaaa: In the next iterations I am going to bring everything to the same form

Lainaaa: But before, this and "send screen" should be merged

PingPongEzZ (dismissed):

PR #742, opened 2025-03-19
[trello.com/c/QCnKo03A] Pop-up redesign

PingPongEzZ:


Adamant/Modules/Chat/View/ChatViewController.swift
```diff
@@ -1066,6 +1072,13 @@ private extension ChatViewController {
         navigationController?.pushViewController(vc, animated: true)
     }
     
+    func didTapShowTimeSettings() {
+        if let appSettings = URL(string: "App-prefs:SCREEN_TIME"),
+            UIApplication.shared.canOpenURL(appSettings) {
```
Lainaaa, line 1081: What about case when it is not able to open?

Adamant/Modules/Chat/View/Managers/ChatDialogManager.swift
```diff
@@ -213,6 +215,29 @@ private extension ChatDialogManager {
         dialogService.present(alert, animated: true, completion: nil)
     }
     
+    func showTimestampAlert() {
+        let alert = UIAlertController(
```
Lainaaa: Consider to use AdamantDialogService

Lainaaa: Consider to use AdamantDialogService

vovameister: ++

Adamant/Modules/Chat/View/ChatViewController.swift
PingPongEzZ: It will not happend

PingPongEzZ: I mean it will always be openned

PingPongEzZ: That's a great u've noticed this. I remembered about the macos 'cause of it. thx

art-divin (approved):

vovameister (approved):

PR #743, opened 2025-03-19
Draft - [trello.com/c/Uh04lfmQ] Set adm nodes "select fastest node"

PingPongEzZ:
Set adm nodes "select fastest node" as true to be the default


Lainaaa:
Check the Trello card, requirements were updated.

art-divin (approved):

PR #746, opened 2025-03-20
[trello.com/c/HwvnDzgu] Fix crash for paging view. Improve current paging index calculation. Refactoring of VisibleWalletsService. 

Lainaaa:
Fix crash for paging view:
Before, it pushed selecting coin even when it is no cells on the screen or when the selected index bigger than cells count – I held this cases.

Improve current paging index calculation:
Before it does not hold case when the current index not the last and you remove one coin before, so the index kept the same while it could be decreased by 1 – I held this cases.

Refactoring of VisibleWalletsService:
No more Notification Center from the visible wallets service 🥳

While I was making it I tried to keep in mind about upcoming secret feature 🤪

### DIFF
--- a/Adamant/Modules/Account/AccountViewController/AccountViewController.swift
+++ b/Adamant/Modules/Account/AccountViewController/AccountViewController.swift
@@ -89,10 +89,10 @@ final class AccountViewController: FormViewController {
     
     private var walletViewControllers: [WalletViewController] = []
     
-    private var currentWalletCoinId: String = ""
+    private var currentWalletCoinID: String = ""
     private var currentSelectedWalletItem: WalletCollectionViewCell.Model? {
         viewModel.state.wallets.first { wallet in
-            wallet.coinID == currentWalletCoinId
+            wallet.coinID == currentWalletCoinID
         }
     }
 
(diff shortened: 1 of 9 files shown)


vovameister (approved):
gj!

Adamant/Modules/Account/AccountViewController/AccountViewController.swift
```diff
@@ -982,6 +982,14 @@ final class AccountViewController: FormViewController {
             refreshControl.endRefreshing()
         }
     }
+    
+    private func selectCurrentWallet() {
+        if let index = viewModel.state.wallets.firstIndex(where: { $0.coinID == currentWalletCoinId }) {
+            pagingViewController.select(index: index, animated: false)
+        } else if viewModel.state.wallets.count > 0 {
+            currentWalletCoinId = viewModel.state.wallets.first!.coinID
```
art-divin: Let's remove force unwrapping to be even more cautions, what do you think @Lainaaa?

```swift
else if let firstWalletID = viewModel.state.wallets.first?.coinID {
  currentWalletCoinID = firstWalletID
}
```

Adamant/Modules/Account/AccountViewController/AccountViewController.swift
```diff
@@ -88,13 +89,13 @@ final class AccountViewController: FormViewController {
     
     private var walletViewControllers: [WalletViewController] = []
     
-    private var currentWalletIndex: Int = .zero
+    private var currentWalletCoinId: String = ""
```
art-divin: Suggestion: `Id -> ID`

Adamant/Modules/Account/AccountViewController/AccountViewController.swift
Lainaaa: Thanks!

art-divin (approved):

PR #748, opened 2025-03-20
[trello.com/c/NOMM1JYt] fix avatar flickering

vovameister:
the previous method for some reason causes flickering when updating one cell, I used configureCell that update Date too but without extra animation

https://github.com/user-attachments/assets/d8df01d6-9ec8-4196-b623-8c9639c47b6f


Lainaaa (approved):
If we are talking about cell flickering - it is fixed, but it also contains other animation bugs (they are also persist in master)

art-divin (approved):

PR #751, opened 2025-03-21
 [trello.com/c/hG3JTo2a] Updating chats drops current chat highlighting on macOS

vovameister:
fix bug + optimized the update algorithm

https://github.com/user-attachments/assets/9120c587-4bdc-4895-821b-4096b6616109

https://github.com/user-attachments/assets/5ae0fca1-4431-4752-a6cd-64b2f69286aa

### DIFF
--- a/Adamant/Modules/ChatsList/ChatListViewController.swift
+++ b/Adamant/Modules/ChatsList/ChatListViewController.swift
@@ -198,8 +198,11 @@ final class ChatListViewController: KeyboardObservingViewController {
     
     override func viewWillAppear(_ animated: Bool) {
         super.viewWillAppear(animated)
-        if let indexPath = tableView.indexPathForSelectedRow {
-            tableView.deselectRow(at: indexPath, animated: animated)
+        
+        if UIDevice.current.userInterfaceIdiom == .phone {
+            if let indexPath = tableView.indexPathForSelectedRow {
+                tableView.deselectRow(at: indexPath, animated: animated)
+            }
         }
     }
     override func viewDidLayoutSubviews() {
@@ -361,7 +364,7 @@ final class ChatListViewController: KeyboardObservingViewController {
     
     private func updateUITitles() {
         updatingIndicatorView.updateTitle(title:  String.adamant.chatList.title)
-        tableView.reloadData()
+        tableView.reloadDataPreservingSelection()
         searchController?.searchBar.placeholder = String.adamant.chatList.searchPlaceholder
     }
     
@@ -412,7 +415,7 @@ final class ChatListViewController: KeyboardObservingViewController {
         areMessagesLoaded = true
         performOnMessagesLoadedActions()
         setIsBusy(!synced)
-        tableView.reloadData()
+        tableView.reloadDataPreservingSelection()
     }
     
     // MARK: IB Actions
@@ -483,7 +486,7 @@ final class ChatListViewController: KeyboardObservingViewController {
         
         switch result {
         case .success:
-            tableView.reloadData()
+            tableView.reloadDataPreservingSelection()
         case .failure(let error):
             dialogService.showRichError(error: error)
         }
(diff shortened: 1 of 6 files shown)


Adamant/Modules/ChatsList/ChatListViewController.swift
```diff
@@ -564,6 +568,9 @@ extension ChatListViewController: UITableViewDelegate, UITableViewDataSource {
         return UIView()
     }
     
+    func tableView(_ tableView: UITableView, didDeselectRowAt indexPath: IndexPath) {
+        chatDeselectedIndex = indexPath
+    }
```
art-divin, line 576: Nit: a missing \r\n here

PingPongEzZ (approved):

PingPongEzZ:
Looks like there's a quite big performance boost due to updating of the single cell instead of all table view. will the chatroom change its position on messages sent as it could be? Is this optimization made for preserved messages only?

vovameister:
> Looks like there's a quite big performance boost due to updating of the single cell instead of all table view. will the chatroom change its position on messages sent as it could be? Is this optimization made for preserved messages only?

only for preserved messages

Adamant/Modules/ChatsList/ChatListViewController.swift
```diff
@@ -719,16 +730,15 @@ extension ChatListViewController {
...
     }
     
     @MainActor
     private func loadNewChats(offset: Int) {
         loadNewChatTask = Task {
             await chatsProvider.getChatRooms(offset: offset)
             isBusy = false
-            tableView.reloadData()
-        }
+            tableView.reloadDataPreservingSelection()        }
```
Lainaaa: Move `}` on the next line pls

Adamant/Modules/ChatsList/ChatListViewController.swift
```diff
@@ -1595,4 +1605,12 @@ private extension UITableView {
         reloadRows(at: indexPaths, with: .none)
         selectRow(at: selectedRowIndexPath, animated: false, scrollPosition: .none)
     }
+    
+    func reloadDataPreservingSelection() {
```
Lainaaa, line 1610: Do you have cases when selection should be not preserving?

I mean maybe just override reload data? It becomes simpler for the future because the new dev for example or the old one can forget it and invoke `reloadData` instead and bug will be returned

Adamant/ServiceProtocols/ChatPreservation.swift
```diff
@@ -32,63 +32,68 @@ final class ChatPreservation: ChatPreservationProtocol, @unchecked Sendable {
...
         preservedReplayMessage = [:]
         preservedFiles = [:]
+        
         updateNotifier.send()
     }
-    
-    func preserveMessage(_ message: String, forAddress address: String) {
-        preservedMessages[address] = message
-        updateNotifier.send()
+    func preserveChatState(
```
Lainaaa, line 39: Add blank line pls

Lainaaa (changes requested):

Adamant/Modules/ChatsList/ChatListViewController.swift
vovameister: have have case when it doesnt matter and we can just use reloadData(). I dont think we should override reloadData, just be careful with use it.

Lainaaa: No, i mean cases when it is matter to not preserve? 
if it always does not matter I insist to override it. As a developer will work in this file he will not care about this case and he will just call "reloadData()" while it will return same bug
The probability of human error becomes very high

vovameister: I don't like this approach, when you start working with code written before you, you are required to first of all study the approaches that were used before you. Overriding methods on the contrary makes the code less readable

Lainaaa: I like this quote:
«Design your APIs in a way that actively steers the user toward doing the “right thing” (Xiaodi Wu). Make it hard for programmers to shoot themselves in the foot.»

Excerpt from the book
Advanced Swift
Chris Eidhof

Lainaaa (approved):
Everything good

Consider https://github.com/Adamant-im/adamant-iOS/pull/751/files#r2009465587

PR #752, opened 2025-03-22
[trello.com/c/B9LfGpfD] Fix of a updating state of chats. 

PingPongEzZ:
Fix of a updating state of chats. Added new logic for alert

### DIFF
--- a/Adamant/Helpers/UserDefaultsManager.swift
+++ b/Adamant/Helpers/UserDefaultsManager.swift
@@ -1,0 +1,17 @@
+//
+//  UserDefaultsManager.swift
+//  Adamant
+//
+//  Created by Sergei Veretennikov on 22.03.2025.
+//  Copyright © 2025 Adamant. All rights reserved.
+//
+
+import CommonKit
+
+struct UserDefaultsManager {
+    @UserDefaultsStorage(.needToShowNoActiveNodesAlert) static var needToShowNoActiveNodesAlert: Bool?
+    
+    static func setInitialUserDefaults() {
+        needToShowNoActiveNodesAlert = true
+    }
+}
--- a/Adamant/Helpers/UserDefaultsManager.swift
+++ b/Adamant/Helpers/UserDefaultsManager.swift
@@ -9,9 +9,9 @@
 import CommonKit
 
 struct UserDefaultsManager {
-    @UserDefaultsStorage(.needToShowNoActiveNodesAlert) static var needToShowNoActiveNodesAlert: Bool?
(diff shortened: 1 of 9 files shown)


art-divin:
I have tested and seems like the chat selection still disappears when loading more chats

Adamant/Helpers/UserDefaultsManager.swift
```diff
@@ -0,0 +1,17 @@
...
+//  Copyright © 2025 Adamant. All rights reserved.
+//
+
+import CommonKit
+
+struct UserDefaultsManager {
+    @UserDefaultsStorage(.needToShowNoActiveNodesAlert) static var needToShowNoActiveNodesAlert: Bool?
+    
+    static func setInitialUserDefaults() {
+        needToShowNoActiveNodesAlert = true
```
art-divin: need -> needs

art-divin (approved):

CommonKit/Sources/CommonKit/Core/UserDefaultsWrapper.swift
```diff
@@ -0,0 +1,26 @@
...
+@propertyWrapper
+public struct UserDefaultsStorage<T> {
+    private let defaults = UserDefaults.standard
+    private let key: String
+    
+    public var wrappedValue: T? {
+        get {
+            defaults.object(forKey: key) as? T
+        } set {
+            defaults.set(newValue, forKey: key)
```
Lainaaa: What about case if I want to remove value? If we would use `nil` here it just update the value, but does not remove it from the UserDefaults, right?

PingPongEzZ: Just set as nil :)

PingPongEzZ: And the value would be the same as if there is no value, and it hasn't ever been sat

Lainaaa: But it will store a key, right?

i suggest to handle this case
Something like 
```
if newValue != nil{
    defaults.set(newValue, forKey: key)
}else{
    remove it
}
```

PingPongEzZ: K dude :))

PingPongEzZ: But it will not store anything :)

vovameister (approved):

PR #754, opened 2025-03-23
[trello.com/c/oOCxsoA5] Test pipeline integration

art-divin:
This PR adds GitHub action to run tests for every merge request


Lainaaa (commented):
Should be good, but i see you used "iPhone 16", i think we should be sure that is no additional different logic on MacOS version 

What do you think about it?

vovameister (approved):

vovameister (commented):
Снимок экрана 2025-03-24 в 22 12 26
is it fine

art-divin:
Closing this one in favour of #768

PR #755, opened 2025-03-24
[trello.com/c/iGILackD] ui update

vovameister:
Button style, padding and macOs hover animation

https://github.com/user-attachments/assets/e25ba4df-2fdf-489f-bdbd-34df9bc1acd5

### DIFF
--- a/Adamant/Modules/Chat/View/ChatViewController.swift
+++ b/Adamant/Modules/Chat/View/ChatViewController.swift
@@ -528,17 +528,17 @@ private extension ChatViewController {
     }
     
     func configureLayout() {
-        view.addSubview(scrollToUnreadReactButton)
         view.addSubview(scrollDownButton)
+        view.addSubview(scrollToUnreadReactButton)
         scrollDownButton.snp.makeConstraints {
             $0.trailing.equalToSuperview().inset(scrollDownButtonInset)
             $0.bottom.equalTo(inputBar.snp.top).offset(-scrollDownButtonInset)
             $0.size.equalTo(scrollButtonHeight)
         }
         scrollToUnreadReactButton.snp.makeConstraints {
-            $0.trailing.equalToSuperview().inset(scrollDownButtonInset)
+            $0.centerX.equalTo(scrollDownButton.snp.centerX)
             self.scrollToUnreadBottomConstraint = $0.bottom.equalTo(scrollDownButton.snp.bottom).constraint
-            $0.size.equalTo(scrollButtonHeight)
+            $0.size.equalTo(scrollButtonHeight + 6)
         }
         
         view.addSubview(loadingView)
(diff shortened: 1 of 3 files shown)


CommonKit/Sources/CommonKit/Helpers/UIHelpers/UIColor+adamant.swift
```diff
@@ -26,6 +26,8 @@ extension UIColor {
         public static let success = #colorLiteral(red: 0.2102436721, green: 0.8444728255, blue: 0.6537195444, alpha: 1) //#32d296
         public static let warning = #colorLiteral(red: 0.9622407556, green: 0.4130832553, blue: 0.5054324269, alpha: 1) //#f0506e
         public static let inactive = #colorLiteral(red: 0.5025414228, green: 0.5106091499, blue: 0.5218499899, alpha: 1) //#6d6f72
+        public static let imageBackground = #colorLiteral(red: 0.8470588235, green: 0.8470588235, blue: 0.8470588235, alpha: 1)
```
Lainaaa, line 29: Where did you take it? As i understand, should be used colors that was used before, is not it?

vovameister: I need to make it the same color as image in asset that we have

Lainaaa: Yesterday we discussed this. The previous icon has color by default while now he uses the system one and he needs to set it by his self. However, to be sure i suggest to set the color for the first icon too, to avoid situations where one icon is different from other.

Adamant/Modules/Chat/View/ChatViewController.swift
```diff
@@ -1206,8 +1206,8 @@ extension ChatViewController {
 }
 
 private let scrollToUnreadInset: CGFloat = 10
-private let scrollDownButtonInset: CGFloat = 10
+private let scrollDownButtonInset: CGFloat = 20
 private let messagePadding: CGFloat = 12
 private let filesToolbarViewHeight: CGFloat = 140
 private let targetYOffset: CGFloat = 20
-private let scrollButtonHeight: CGFloat = 38
+private let scrollButtonHeight: CGFloat = 30
```
Lainaaa, line 1213: Why do you decided to change this value while above you use it highet:
```
$0.size.equalTo(scrollButtonHeight + 6)
```

Lainaaa (approved):

Adamant/Modules/Chat/View/ChatViewController.swift
vovameister: the custom image visually takes up more space than SF Symbol

CommonKit/Sources/CommonKit/Helpers/UIHelpers/UIColor+adamant.swift
vovameister: it is impossible to set the color for the first assets image because there are 2 colors
and also its extra code

PingPongEzZ (approved):

CommonKit/Sources/CommonKit/Helpers/UIHelpers/UIColor+adamant.swift
Lainaaa: > it is impossible to set the color for the first assets image because there are 2 colors

wdym

Adamant/Modules/Chat/View/ChatViewController.swift
Lainaaa: I think it because SF symbol != image. Symbol height != line height while image height == line height. 
If you could find better approach than just visual feelings it would be great!

PR #758, opened 2025-03-25
[trello.com/c/aZtAfZ44] add read message entering chat

vovameister:


Adamant/Modules/Chat/View/ChatViewController.swift
```diff
@@ -172,6 +172,7 @@ final class ChatViewController: MessagesViewController {
         inputBar.isUserInteractionEnabled = true
         chatMessagesCollectionView.fixedBottomOffset = nil
         
+        updateUnreadMessages()
```
Lainaaa, line 175: Instead of checking it before view appears maybe the better way is to make it when the view controller will be inited?

Lainaaa (approved):
It should work

Adamant/Modules/Chat/View/ChatViewController.swift
vovameister: viewDidAppear is not before appear its after

Lainaaa: Okay, my question still the same

> Instead of checking it after view appears maybe the better way is to make it when the view controller will be inited?

PingPongEzZ: Updating may be a tough task and It might be uncompleted at the moment of the screen presentation. Better to do it syncronuosely on init as Lainaaa said

vovameister: we need visible indexes from collection view for update,  how we can have it in init?

PingPongEzZ (approved):

PR #759, opened 2025-03-26
[trello.com/c/YsjIF3DY] (Secret wallets) Handle logout/login case

Lainaaa:
Main logic is to subscribe on notification event for logout and to remove all secret wallets when logout happens. After that other services receive a new state because they are state subscribers.


Adamant/Services/SecretWalletsManager.swift
```diff
@@ -76,4 +76,10 @@ final class AdamantSecretWalletsManager: SecretWalletsManagerProtocol {
         state.currentWallet = state.regularWallet
         statePublisher.send(state)
     }
+    
+    func removeAllSecretWallets() {
+        lock.lock()
```
vovameister: are you sure that we need lock here?

PingPongEzZ: We have an atomic wrapper

Lainaaa: > are you sure that we need lock here?

@vovameister I think it should be everywhere where state could be mutated

Lainaaa: > We have an atomic wrapper

@PingPongEzZ Yes, if I remember right i even tried to use it but something was wrong. If you think that is really necessary to use here I can to give it the second chance, however, not in this PR but in the main one - [trello.com/c/vawidi4o](https://github.com/Adamant-im/adamant-iOS/tree/trello.com/c/vawidi4o)

vovameister (commented):
please write а description

vovameister (approved):

PR #764, opened 2025-03-28
[trello.com/c/5yQDEmmD] Update icons for iOS 18

Lainaaa:


vovameister (commented):
are you sure that we dont need icon that you deleted anymore?

PingPongEzZ (approved):

art-divin (approved):

Lainaaa:
@vovameister Yes, you have 2 options:
* you can provide full pack of icons for different data sets (MacOS, iPad and etc) 
* you can just give an 1024x1024 icon and others Xcode will create automatically

Before was the first one while now i used the second

PR #761, opened 2025-03-27
[trello.com/c/txGIRwKv] Adamant-wallets parsing implementation

PingPongEzZ:
Here's the implementation of merging of our wallets contract in adamant-wallets. 
Now you can access any coin or specific token on the chain or the list of tokens on the blockchain by `AnyTokensStorage`'s subscripts.
image

PR #766, opened 2025-03-28
Draft - [trello.com/c/Tz5VHyL4] Implementation of tokens data preview 

PingPongEzZ:
https://github.com/user-attachments/assets/012f6569-9a9a-410a-896b-b224633991cf

### DIFF
--- a/Adamant/Modules/Settings/NodesAndTokensInfo/CoinInfoDTOViewController.swift
+++ b/Adamant/Modules/Settings/NodesAndTokensInfo/CoinInfoDTOViewController.swift
@@ -1,0 +1,131 @@
+//
+//  CoinInfoDTOViewController.swift
+//  Adamant
+//
+//  Created by Sergei Veretennikov on 28.03.2025.
+//  Copyright © 2025 Adamant. All rights reserved.
+//
+
+import AdamantWalletsKit
+import UIKit
+import CommonKit
+
+private protocol AnyPrimitive: Sendable {}
+
+extension Int: AnyPrimitive {}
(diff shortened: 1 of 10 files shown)


PingPongEzZ:
Draft until https://github.com/Adamant-im/adamant-iOS/pull/761 mergred

Adamant/Modules/Settings/NodesAndTokensInfo/CoinInfoDTOViewController.swift
```diff
@@ -0,0 +1,131 @@
...
+        tableView.register(UITableViewCell.self, forCellReuseIdentifier: "cell")
+        tableView.dataSource = self
+        tableView.delegate = self
+        tableView.reloadData()
+    }
+    
+    private func setDataSource(for coinInfo: CoinInfoDTO) {
+        guard let data = try? JSONEncoder().encode(coinInfo) else { return }
+
+        func flatten(value: Any, prefix: String) {
```
art-divin, line 54: It seems that this code should be outside of the ViewController, a separate file for example and entity as well. Can you simply move this one out @PingPongEzZ ?

art-divin (approved):

Adamant/Modules/Settings/NodesAndTokensInfo/CoinInfoDTOViewController.swift
PingPongEzZ: I've put it here because this algo needed only in the scope of this function not all the class even in internal visibility of it

PR #767, opened 2025-03-31
[trello.com/c/jKqpxiCa] update new message line and chat entering scroll

vovameister:
- Change padding and color of new message line
- now separator not adding after entering chat for new messages, only while entering if we have some new messages
- Now the scroll priority when entering the chat is as follows: 
push notifications taped message -> newmessages separator -> save position
P.s. I cant put everything in one place like in updateMessagesPosition() because scroll to push notifications taped message sometimes need extra message loading request. So its separate.


Lainaaa (approved):
lgtm

PingPongEzZ (approved):

PR #773, opened 2025-04-01
[trello.com/c/TtOBs0dQ] Fix: Markdown doesn't support headers in Chat list

Lainaaa:
The "You" prefix was broking markdown header logic. Now i paste it after markdown parsing while earlier it was before. 

This fix does not hold case when you tap "reply" and preview is appeared, in this case on the preview will be String without markdown symbols, instead of markdown because it does not support markdown for now.


vovameister (commented):
need also add support for Draft (next method private func shortDescription(for address: String) -> NSAttributedString?)

vovameister (approved):

PR #776, opened 2025-04-01
[trello.com/c/lwOnjxPQ] fix scroll if new messages not on a full screen

vovameister:
added a method that sums the height of new cells until it exceeds the height of the visible table field and only then adds an extra offset if need

### DIFF
--- a/Adamant/Modules/Chat/View/ChatViewController.swift
+++ b/Adamant/Modules/Chat/View/ChatViewController.swift
@@ -45,7 +45,7 @@ final class ChatViewController: MessagesViewController {
     private var scrollToUnreadBottomConstraint: Constraint?
     private var isScrollDownButtonHidden = true
     private var previousUnreadCount: Int = 0
-    private var scrollToPosition: UICollectionView.ScrollPosition = .top
+    private var scrollToPositionType: UICollectionView.ScrollPosition = .top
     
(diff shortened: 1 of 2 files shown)


art-divin (approved):

Adamant/Modules/Chat/View/ChatViewController.swift
```diff
@@ -950,12 +950,14 @@ private extension ChatViewController {
...
             messagesCollectionView.scrollToItem(
                 at: .init(item: .zero, section: index),
-                at: scrollToPosition,
+                at: scrollToPositionType,
                 animated: animated
             )
             
             if setExtraOffset {
-                setExtraOffsetForNewMessages()
+                if checkIfNeedExtraOffsetForUnreadMessages() {
```
Lainaaa, line 958: Tip: you could write ifs on the same line

Adamant/Modules/Chat/View/ChatViewController.swift
```diff
@@ -985,6 +987,30 @@ private extension ChatViewController {
         messagesCollectionView.setContentOffset(newOffset, animated: false)
     }
     
+    func checkIfNeedExtraOffsetForUnreadMessages() -> Bool {
```
Lainaaa, line 990: isExtraOffsetForUnreadMessagesNeeds

PR #784, opened 2025-04-06
[trello.com/c/AvS5QpYb] Bug: Warning window "No active nodes" on Pull-to-Refresh on Account screen

Lainaaa:
Now the protocol has two methods related to updates, plus a private method that performs the actual update.

I considered two options:
* Update the existing method and include the UI balance update logic there
* Create a new dedicated method

I chose the second option to avoid excessive refactoring — hope that makes sense to you too!
If you have any suggestions, especially around naming, I’m open to improvements.


vovameister (approved):

Adamant/Modules/Wallets/ERC20/ERC20WalletService.swift
```diff
@@ -255,8 +255,14 @@ final class ERC20WalletService: WalletCoreProtocol, ERC20GasAlgorithmComputable,
         }
     }
 
+    func updateWithRefreshUIBalance(){
```
PingPongEzZ: Fix spaces between the braces in the MR

Lainaaa: I fixed in this case, but linter was removed and it is harder to recognize this warnings. 
However, @art-divin  said that we don't need to worry about formatting anymore because we have auto-format feature.

PR #785, opened 2025-04-06
Optimize the carousel updating. Instead of updating the whole carousel update only inner cells.

Lainaaa:
* Optimized paging reload: instead of reloading the entire page, only the relevant cells are updated. This should help fix white screen bugs, crashes, and improve behavior with secret wallets.
* Preserved the current wallet index to show the appropriate message when nodes are unavailable.
* Removed any manual index control logic related to the carousel.


Adamant/Helpers/UIImage+Comparable.swift
```diff
@@ -0,0 +1,18 @@
...
+//  CommonKit
+//
+//  Created by Dmitrij Meidus on 06.04.25.
+//
+
+import UIKit
+
+extension UIImage: @retroactive Comparable {
+    public static func < (lhs: UIImage, rhs: UIImage) -> Bool {
+        fatalError()
```
Lainaaa: I think this approach works well — it enables comparison of images using ==, while explicitly disallowing ordering with <.

Curious to hear your thoughts — do you think it makes sense, or would you approach it differently?

Adamant.xcodeproj/project.pbxproj
```diff
@@ -31,7 +31,7 @@
 		A5241B70262DEDE1009FA43E /* Clibsodium in Frameworks */ = {isa = PBXBuildFile; productRef = A5241B6F262DEDE1009FA43E /* Clibsodium */; };
 		A5241B77262DEDEF009FA43E /* Clibsodium in Frameworks */ = {isa = PBXBuildFile; productRef = A5241B76262DEDEF009FA43E /* Clibsodium */; };
 		A5241B7E262DEDFE009FA43E /* Clibsodium in Frameworks */ = {isa = PBXBuildFile; productRef = A5241B7D262DEDFE009FA43E /* Clibsodium */; };
-		A530B0D82842110D003F0210 /* BuildFile in Frameworks */ = {isa = PBXBuildFile; };
+		A530B0D82842110D003F0210 /* (null) in Frameworks */ = {isa = PBXBuildFile; };
```
PingPongEzZ, line 34: Looks like something wrong here

Adamant/Modules/Account/AccountViewController/AccountViewController.swift
```diff
@@ -90,13 +90,10 @@ final class AccountViewController: FormViewController {
...
-    private var currentWalletCoinID: String = ""
-    private var currentSelectedWalletItem: WalletCollectionViewCell.Model? {
-        viewModel.state.wallets.first { wallet in
-            wallet.coinID == currentWalletCoinID
-        }
-    }
+    
+    private lazy var currentSelectedWallet: AccountWalletCellState? = { viewModel.state.wallets.first(where: { wallet in
+        wallet.model.index == 0
+    })}()
```
PingPongEzZ: Format please a property

Adamant/Modules/Account/AccountViewController/AccountWallets/AccountWalletsViewModel.swift
```diff
@@ -43,25 +43,29 @@ extension AccountWalletsViewModel {
...
     fileprivate func updateInfo(for wallet: WalletService) {
         let coreService = wallet.core
-        if let index = state.wallets.firstIndex(where: { $0.coinID == coreService.tokenUniqueID }) {
-            state.wallets[index].balance = coreService.wallet?.balance ?? 0
-            state.wallets[index].isBalanceInitialized = coreService.wallet?.isBalanceInitialized ?? false
-            state.wallets[index].notificationBadgeCount = coreService.wallet?.notifications ?? 0
+        let tokenID = coreService.tokenUniqueID
+        if let index = state.wallets.firstIndex(where: { $0.model.coinID == tokenID }) {
+            let cellState = state.wallets[index]
+            var newModel = cellState.model
```
PingPongEzZ: Notice that here's some ambigious property naming. The code in the `else` block are truly creating a `newModel`, but in `if` block you are getting an already existed model, so that should be an `existedModel` I suppose

Adamant.xcodeproj/project.pbxproj
art-divin: Hey,

it is an artefact of either Xcode or most likely cocoapods; should not do any harm unless we use XcodeProj, at that time we would be migrating to Tuist or something. 😄

PingPongEzZ (approved):

PR #788, opened 2025-04-07
[trello.com/c/aedpNhWv] fix avatar blinking

vovameister:
After several hours and hundreds of swipes down I found that mostly it was for two reasons :
1. refreshDatesIfNeeded when we are updating through tableView willDisplay, so now its not updating while scrolling
2. handleRefresh multiple times if you try to spam update sometimes flick avatars
maybe there are some other scenarios, but even those that I fixed were quite rare, without changing the architectural approach it is unlikely that it will be possible to cure this by more than 99%

### DIFF
--- a/Adamant/Modules/ChatsList/ChatListViewController.swift
+++ b/Adamant/Modules/ChatsList/ChatListViewController.swift
@@ -77,6 +77,8 @@ final class ChatListViewController: KeyboardObservingViewController {
         }
     }
     private var chatDeselectedIndex: IndexPath?
+    private var isScrolling = false
+    private var isRefreshing = false
 
     let defaultAvatar = UIImage.asset(named: "avatar-chat-placeholder") ?? .init()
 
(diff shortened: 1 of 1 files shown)


Lainaaa (commented):
I think it is not a good solution  
It makes count of cases when it flickering smaller while we need to remove them altogether
I found that people already discussed this problem, did you try it?
https://stackoverflow.com/questions/15196927/reload-uitableview-with-new-data-caused-flickering
https://developer.apple.com/forums/thread/71287

If nothing was helped, lets discuss about it in a team.

vovameister:
> I think it is not a good solution It makes count of cases when it flickering smaller while we need to remove them altogether I found that people already discussed this problem, did you try it? https://stackoverflow.com/questions/15196927/reload-uitableview-with-new-data-caused-flickering https://developer.apple.com/forums/thread/71287
> 
> If nothing was helped, lets discuss about it in a team.

Try it on your device, if it still reproduce more then 1 to 1000, I will continue to work on this

Adamant/Modules/ChatsList/ChatListViewController.swift
```diff
@@ -615,13 +609,25 @@ extension ChatListViewController: UITableViewDelegate, UITableViewDataSource {
...
         scrollUpButton.isHidden = offsetY < cellHeight * 0.75
     }
+    
     func scrollViewDidEndDragging(_ scrollView: UIScrollView, willDecelerate decelerate: Bool) {
         guard scrollView.contentOffset.y <= 0, scrollView.contentOffset.y < -100 else { return }
 
-        Task {
-            await handleRefresh()
+        if !isRefreshing {
+            isRefreshing = true
```
Lainaaa, line 617: Would be better to mark that it is refreshes in `handleRefresh` method, to control it on one place. Like you did in 
```defer { isRefreshing = false }```

Lainaaa (approved):
I tested, it does not blinking any more

Adamant/Modules/ChatsList/ChatListViewController.swift
vovameister: why force the program to execute unnecessary code

PR #789, opened 2025-04-07
[trello.com/c/sG80IZan] remove chat reeding when app is not active macOs

vovameister:
It has a limitation, if you just remove the application from focus, reading will still continue. You need to switch to another application so that the messenger goes to the background

### DIFF
--- a/Adamant/Modules/Chat/View/ChatViewController.swift
+++ b/Adamant/Modules/Chat/View/ChatViewController.swift
@@ -46,6 +46,7 @@ final class ChatViewController: MessagesViewController {
     private var isScrollDownButtonHidden = true
     private var previousUnreadCount: Int = 0
     private var scrollToPositionType: UICollectionView.ScrollPosition = .top
+    private var isAppActive = true
 
     private lazy var inputBar = ChatInputBar()
     private lazy var loadingView = LoadingView()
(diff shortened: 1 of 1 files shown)


Adamant/Modules/Chat/View/ChatViewController.swift
```diff
@@ -309,10 +310,18 @@ extension ChatViewController {
...
             .sink { @MainActor [weak self] _ in
                 guard let self = self else { return }
+                self.isAppActive = true
+                self.updateUnreadMessages()
                 let indexes = self.messagesCollectionView.indexPathsForVisibleItems
                 self.viewModel.updatePreviewFor(indexes: indexes)
             }
             .store(in: &subscriptions)
+        
+        NotificationCenter.default.publisher(for: UIApplication.willResignActiveNotification)
```
Lainaaa, line 320: What do you think about creating state manager for this? Instead of managering it in every class we can create one place 
It also makes an already overloaded class lighter.

vovameister: I don't think it's worth it considering that we directly receive notification from UIApplication and we don't need to sync anything

art-divin (approved):
Good for a quick fix, but essentially ChatViewController became bigger while test coverage did not increase, consider adding tests if possible

vovameister:
> Good for a quick fix, but essentially ChatViewController became bigger while test coverage did not increase, consider adding tests if possible

unfortunately this case cannot be covered by any type of tests

PR #792, opened 2025-04-09
[trello.com/c/2Dz3alvN] Fixed a memory leak.

PingPongEzZ:
I need to override HealthCheckWrapped because now we have an option to cancel the tasks
 and there is some situations when we are waiting for a connection on the node and recursively calling `request`
 function without any break condition and that is bad and unsafe. I've added a functionality of pushing the associated with the task `UUID`
 and checking it for a cancellation before doing something in the recursive function
 `request(waitsForConnectivity:taskId:isCancelled:_ request:)`


art-divin (approved):
Consider using copilot or cursor AI to generate tests for these conditions. Can be a part of a separate PR.

PR #793, opened 2025-04-10
[trello.com/c/ONt29HWW] Save the index of the current selected crypto in a carousel when switching between secrets and regular.

Lainaaa:
By default, reloadData resets the currently selected item index.
My approach is to re-select the item after the reload.


art-divin (approved):

art-divin:
[{Secret wallets} Save the index of the current selected crypto in a carousel when switching between secrets and regular](https://trello.com/c/ONt29HWW/894-secret-wallets-save-the-index-of-the-current-selected-crypto-in-a-carousel-when-switching-between-secrets-and-regular)

PR #797, opened 2025-04-10
[trello.com/c/iGILackD] many chat updates

vovameister:
1. return scrolling to center using buttons and chat entering  besides scroll to "new messages" separator
2. update hightlight animation, hope I also fix that sometimes its not working or removing to early 
3. add short vibro while highliting
4. remove reading messages while autoscrolling (scroll to item, not scroll to bottom)
5. also remove button animation  while autoscrolling
6. put position restore back so it should be earlier to not reproduce chicken animation
7. return animate = true as it was before for autoscrolling

### DIFF
--- a/Adamant/Modules/Chat/View/ChatScrollState.swift
+++ b/Adamant/Modules/Chat/View/ChatScrollState.swift
@@ -1,0 +1,26 @@
+//
+//  ChatScrollState.swift
+//  Adamant
+//
+//  Created by Владимир Клевцов on 11. 4. 2025..
+//  Copyright © 2025 Adamant. All rights reserved.
+//
+import SnapKit
+
+struct ChatScrollState {
+    var messagesLoaded = false
+    var isScrollPositionNearlyTheBottom = true
+    var viewAppeared = false
+    var scrollToUnreadBottomConstraint: Constraint?
+    var isScrollDownButtonHidden = true
+    var previousUnreadCount: Int = 0
+    var isAnimatingCellHighlight = false
+    var isAutoScrolling = false
+    var isAppActive = true
+    var isScrollingToBottom = false
+    
+    //calculation for animate, might use for something else in the futer
+    var isAllowed: Bool {
+        messagesLoaded && !isAutoScrolling && !isScrollingToBottom
+    }
+}
--- a/Adamant/Modules/Chat/View/ChatViewController.swift
+++ b/Adamant/Modules/Chat/View/ChatViewController.swift
@@ -45,7 +45,8 @@ final class ChatViewController: MessagesViewController {
     private var scrollToUnreadBottomConstraint: Constraint?
     private var isScrollDownButtonHidden = true
     private var previousUnreadCount: Int = 0
-    private var scrollToPositionType: UICollectionView.ScrollPosition = .top
+    private var isAnimatingCellHighlight = false
+    private var isAutoScrolling: Bool = false
 
     private lazy var inputBar = ChatInputBar()
     private lazy var loadingView = LoadingView()
@@ -195,22 +196,6 @@ final class ChatViewController: MessagesViewController {
                 ? nil
                 : chatMessagesCollectionView.bottomOffset
         )
-    }
-
-    override func collectionView(
-        _ collectionView: UICollectionView,
-        willDisplay cell: UICollectionViewCell,
-        forItemAt indexPath: IndexPath
-    ) {
-        // TODO: refactor for architecture
-        if let index = viewModel.needToAnimateCellIndex,
-            indexPath.section == index
-        {
-            cell.isSelected = true
-            cell.isSelected = false
-            viewModel.needToAnimateCellIndex = nil
-        }
-        super.collectionView(collectionView, willDisplay: cell, forItemAt: indexPath)
     }
 
     override func scrollViewDidEndDecelerating(_: UIScrollView) {
@@ -423,11 +408,9 @@ extension ChatViewController {
 
         viewModel.$scrollToIdAndPosition
             .sink { [weak self] in
-                guard let idAndPosition = $0
+                guard let id = $0
                 else { return }
-
-                self?.scrollToPositionType = idAndPosition.position
-                self?.scrollToPosition(.messageId(idAndPosition.id), animated: false, setExtraOffset: self?.scrollToPositionType == .top)
+                self?.scrollToPosition(.messageId(id), animated: true)
                 self?.viewModel.messageIdToShow = nil
             }
             .store(in: &subscriptions)
@@ -571,6 +554,7 @@ extension ChatViewController {
     }
 
     fileprivate func updateUnreadMessages() {
+        guard !isAutoScrolling else { return }
         guard let unreadIndexes = viewModel.unreadMesaggesIndexes, !unreadIndexes.isEmpty else { return }
         let visibleIndexPaths = messagesCollectionView.indexPathsForVisibleItems
 
@@ -765,6 +749,9 @@ extension ChatViewController {
         chatMessagesCollectionView.reloadData(newIds: viewModel.messages.map { $0.id }, isOnBottom: isScrollPositionNearlyTheBottom)
         scrollDownOnNewMessageIfNeeded(previousBottomMessageId: bottomMessageId)
         bottomMessageId = viewModel.messages.last?.messageId
+        if !messagesLoaded {
+            viewModel.startPosition.map { scrollToPosition($0) }
+        }
     }
 
     fileprivate func updateMessagesPosition() {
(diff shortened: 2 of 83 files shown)


Adamant/Modules/Chat/View/ChatViewController.swift
```diff
@@ -45,8 +45,11 @@ final class ChatViewController: MessagesViewController {
     private var scrollToUnreadBottomConstraint: Constraint?
     private var isScrollDownButtonHidden = true
     private var previousUnreadCount: Int = 0
-    private var scrollToPositionType: UICollectionView.ScrollPosition = .top
+    private var isAnimatingCellHighlight = false
+    private var isAutoScrolling = false
     private var isAppActive = true
+    //need separate var to read messages but dont animate button
```
art-divin: ~need~

Adamant/Modules/Chat/View/ChatViewController.swift
```diff
@@ -45,8 +45,11 @@ final class ChatViewController: MessagesViewController {
     private var scrollToUnreadBottomConstraint: Constraint?
```
Lainaaa: Consider to move everything into a separate state?
I think it is more then enough variables here

Adamant/Modules/Chat/View/ChatViewController.swift
```diff
@@ -846,7 +832,7 @@ extension ChatViewController {
         }
         previousUnreadCount = count
         let updateAlpha = { self.scrollToUnreadReactButton.alpha = (count == 0) ? 0 : 1 }
-        if messagesLoaded {
+        if messagesLoaded && !isAutoScrolling && !isScrollingToBottom {
```
Lainaaa: This could be moved to a computed property in the state

Lainaaa (commented):
Consider my comments, I think it becomes more complex and comples, lets make it easier!

Adamant/Modules/Chat/View/ChatViewController.swift
vovameister: would be grate

Adamant/Modules/Chat/View/ChatViewController.swift
vovameister: got it

Adamant/Modules/Chat/View/ChatScrollState.swift
```diff
@@ -0,0 +1,26 @@
...
+//  ChatScrollState.swift
+//  Adamant
+//
+//  Created by Владимир Клевцов on 11. 4. 2025..
+//  Copyright © 2025 Adamant. All rights reserved.
+//
+import SnapKit
+
+struct ChatScrollState {
+    var messagesLoaded = false
```
Lainaaa: isMessagesLoaded

Adamant/Modules/Chat/View/ChatScrollState.swift
```diff
@@ -0,0 +1,26 @@
...
+    var isScrollPositionNearlyTheBottom = true
+    var viewAppeared = false
```
Lainaaa: isViewAppeared

Adamant/Modules/Chat/View/ChatScrollState.swift
```diff
@@ -0,0 +1,26 @@
...
+    var scrollToUnreadBottomConstraint: Constraint?
+    var isScrollDownButtonHidden = true
+    var previousUnreadCount: Int = 0
+    var isAnimatingCellHighlight = false
+    var isAutoScrolling = false
+    var isAppActive = true
+    var isScrollingToBottom = false
+    
+    //calculation for animate, might use for something else in the futer
+    var isAllowed: Bool {
```
Lainaaa: isAllowedTo`here what should be allowed`

Adamant/Modules/Chat/View/ChatViewController.swift
```diff
@@ -589,14 +564,15 @@ extension ChatViewController {
...
+        state.scrollToUnreadBottomConstraint?.update(offset: offset)
+        if state.messagesLoaded {
             self.view.layoutIfNeeded()
         }
     }
 
     fileprivate func updateUnreadMessages() {
-        if isMacOS && !isAppActive {
+        guard !state.isAutoScrolling else { return }
+        if isMacOS && !state.isAppActive {
```
Lainaaa, line 575: You could combine this if and guard

Adamant/Modules/Chat/ViewModel/ChatViewModel.swift
```diff
@@ -89,7 +89,7 @@ final class ChatViewModel: NSObject {
     let minIndexForStartLoadNewMessages = 4
     let minOffsetForStartLoadNewMessages: CGFloat = 100
     var tempOffsets: [String] = []
-    var needToAnimateCellIndex: Int?
+    var needToAnimateCellIndex: String?
```
Lainaaa: It sounds like a bool

Adamant/Modules/Chat/ViewModel/ChatViewModel.swift
```diff
@@ -134,21 +134,14 @@ final class ChatViewModel: NSObject {
     @ObservableValue private(set) var dateHeaderHidden: Bool = true
     @ObservableValue var inputText = ""
     @ObservableValue var replyMessage: MessageModel?
-    @ObservableValue var scrollToIdAndPosition: (id: String, position: UICollectionView.ScrollPosition)?
+    @ObservableValue var scrollToIdAndPosition: String?
```
Lainaaa, line 138: Could be this also moved to the new state?

Adamant/Modules/Chat/ViewModel/ChatViewModel.swift
```diff
@@ -265,6 +259,18 @@ final class ChatViewModel: NSObject {
         }
     }
 
+    func makeStartPosition() {
```
Lainaaa: private?

Lainaaa: Instead of creating a separate func you can calculate this on variable


```
    lazy var startPosition: ChatStartPosition? = {
        guard messageIdToShow == nil,
              let address = chatroom?.partner?.address else {
            startPosition = nil
            return nil
        }
        
        return chatsProvider
            .getChatPositon(for: address)
            .map { .offset(.init($0)) }
    }()
```

Just want to show another approach

Lainaaa (commented):
Consider moving some of the viewModels variables to the new state

Also, I don't like approach to combine dialog and vibro, instead - use vibro directly if needs

Lainaaa (commented):
Also consider moving state to the viewModel

Adamant/Modules/Chat/ViewModel/ChatViewModel.swift
vovameister: no reason to lazy because its gonna be checked every time, and I also need to set it once

Adamant/Modules/Chat/View/ChatViewController.swift
vovameister: I think now it looks cleaner

Adamant/Modules/Chat/View/ChatViewControllerState.swift
```diff
@@ -0,0 +1,26 @@
...
+    var scrollToUnreadBottomConstraint: Constraint?
+    var isScrollDownButtonHidden = true
+    var previousUnreadCount: Int = 0
+    var isAnimatingCellHighlight = false
+    var isAutoScrolling = false
+    var isAppActive = true
+    var isScrollingToBottom = false
+    
+    //calculation for animate, might use for something else in the futer
+    var isAimateAllowed: Bool {
```
Lainaaa: Aim🫡

Lainaaa (dismissed):

Adamant/Modules/Chat/ViewModel/ChatViewModel.swift
```diff
@@ -1092,6 +1085,10 @@ extension ChatViewModel {
     func unredMessageCount() -> Int? {
         unreadMessagesIds?.count
     }
+    
+    func shortVibro() {
```
Lainaaa, line 1089: You don't need to create additional method for this

Adamant/Modules/Chat/View/ChatViewControllerState.swift
```diff
@@ -0,0 +1,26 @@
...
+    var isViewAppeared = false
+    var scrollToUnreadBottomConstraint: Constraint?
+    var isScrollDownButtonHidden = true
+    var previousUnreadCount: Int = 0
+    var isAnimatingCellHighlight = false
+    var isAutoScrolling = false
+    var isAppActive = true
+    var isScrollingToBottom = false
+    
+    //calculation for animate, might use for something else in the futer
```
art-divin: ~animate~ -> animation, ~futer~ -> future

art-divin (approved):

PR #807, opened 2025-04-13
[trello.com/c/aedpNhWv] move avatar generate and set to main thread

vovameister:
Because the avatars were generated in the global queue and assigned in a different runloop than the rest of the cell, there were visual flickering of the avatars. I test on my iphone 15 pro and in debug each avatar was created for like 0.002 sec, so it should be ok to do it on main queue


Lainaaa (approved):
What about previous changes? Should we Roll it back?

Lainaaa (approved):
Great work!

PR #809, opened 2025-04-14
[trello.com/c/QBfqpo6A] Store A:B ratio and use it as default.

Lainaaa:


Lainaaa:
[Feat/UI: Allow more side panel width on macOS](https://trello.com/c/QBfqpo6A/892-feat-ui-allow-more-side-panel-width-on-macos)

art-divin (approved):

PR #815, opened 2025-04-15
[trello.com/c/7IANJAJA] Setting new snipper logic

PingPongEzZ:
I've set a new snipper logic and fixed few bugs here:
Setting IPFS from hardcode to adamant-wallets
Setting an logout functionality to be synchronious

### DIFF
--- a/Adamant/Services/DataProviders/AnyChatUpdatingNotificator.swift
+++ b/Adamant/Services/DataProviders/AnyChatUpdatingNotificator.swift
@@ -1,0 +1,67 @@
+//
+//  AnyChatUpdatingNotificator.swift
+//  Adamant
+//
+//  Created by Sergei Veretennikov on 15.04.2025.
+//  Copyright © 2025 Adamant. All rights reserved.
+//
+
+import CommonKit
+import Foundation
+
+protocol AnyChatUpdatingNotificator {
+    var needToShowChatUpdateSpinnerObeserver: AnyObservable<Bool> { get }
+    
+    func updateLastUpdateTime()
+    func startMonitoring()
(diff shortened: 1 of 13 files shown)


adamantmm (approved):

art-divin:
[{Connection}: Update spinner algorithm for Chat & Chat list screens + Health-check optimization](https://trello.com/c/7IANJAJA/889-connection-update-spinner-algorithm-for-chat-chat-list-screens-health-check-optimization)

Adamant/Services/DataProviders/AnyChatUpdatingNotificator.swift
```diff
@@ -0,0 +1,67 @@
...
+    
+    func updateLastUpdateTime()
+    func startMonitoring()
+    func stopMonitoring()
+}
+
+final class ChatUpdatingNotificator: AnyChatUpdatingNotificator {
+    @ObservableValue private var needToShowChatUpdateSpinner: Bool = false
+    private var lastUpdate: Date?
+    private var continuation: AsyncThrowingStream<Bool, Error>.Continuation?
```
Lainaaa, line 23: You don't handle throwing so you could use could you use just `AsyncThrowingStream`?

Lainaaa (approved):
lgtm

Adamant/Services/DataProviders/AnyChatUpdatingNotificator.swift
PingPongEzZ: Just in case. Probably there is will not be a situation when the error occurs from the stream. But it is not principal which of the continuations you can use.

PR #817, opened 2025-04-16
[trello.com/c/iGILackD] animate reaction and message separate

vovameister:
- add opponent reaction animation
- if the scroll goes to the reaction (including in app push notifications), the reaction is animated, the reaction itself without the message.  
- Now all cell animation is done through the overlay
- when scrolling down there is no longer a selection of the message and vibration


Adamant/Modules/Chat/ViewModel/ChatViewModel.swift
```diff
@@ -91,6 +91,7 @@ final class ChatViewModel: NSObject {
     let minOffsetForStartLoadNewMessages: CGFloat = 100
     var tempOffsets: [String] = []
     var cellIdForAnimation: String?
+    var animationType: MessageAnimationType?
```
Lainaaa, line 93: Is not it better to store it in a controller?

art-divin (approved):

PR #818, opened 2025-04-16
[trello.com/c/4QfDaVMi] Excessive token standard for ADM, BTC, ETH, DOGE, DASH, KLY

Lainaaa:
Handle case for ETH which has currencySymbol different from the currencyNetwork while it still not a token.


Lainaaa:
[Bug: Excessive token standard for ADM, BTC, ETH, DOGE, DASH, KLY](https://trello.com/c/4QfDaVMi/828-bug-excessive-token-standard-for-adm-btc-eth-doge-dash-kly)

vovameister (approved):

PR #827, opened 2025-04-21
[trello.com/c/46S4GSZ5] task sleep reduced

### LINKED ISSUE #916: [Bug] Hold pressing on a message does not work properly on macOS (Copy action)
### Summary

On macOS, long-press behavior for messages is inconsistent and buggy

### Steps to reproduce

1. Long-press menu appears only for **incoming messages** and doesn’t work for outgoing ones
2. Menu position looks misaligned/buggy
3. **Single tap** unexpectedly copies text (on both iOS and macOS)

### iOS
- **Single tap:** No action / Highlight
- **Long-press:** Show context menu (Reply, Copy, etc.)
- **Two-finger tap:** Copy text directly [Cancelled this behavior because of both poor UX and difficulties in implementation]

### macOS
- **Single tap:** No action / Highlight
- **Long-press (left mouse button):** Copy text (if message contains text)
- **Two-finger tap (trackpad / right-click):** Show context menu

### Actual behavior

- Long-press menu only appears for incoming messages, not outgoing ones
- Menu position is buggy
- Single tap **copies** text (unexpected) — both on iOS and macOS

## Acceptance Criteria
- On iOS:
  - Single tap does nothing
  - Long-press shows menu
  - Two-finger tap copies text
- On macOS:
  - Single tap does nothing
  - Long-press copies text
  - Two-finger tap shows menu

## Notes
- Long-press should always trigger animation for feedback
- Copy action only applies if the message contains text

### Environment

macOS
iOS

vovameister:
from 1,5 sec -> 1 sec

### DIFF
--- a/Adamant/Modules/Chat/View/Subviews/ChatBaseMessage/ChatMessageCell.swift
+++ b/Adamant/Modules/Chat/View/Subviews/ChatBaseMessage/ChatMessageCell.swift
@@ -499,7 +499,7 @@ private extension ChatMessageCell {
             messageContainerView.animatePressDown()
             
             Task { [weak self] in
-                try? await Task.sleep(nanoseconds: UInt64(1.5) * 1_000_000_000)
+                try? await Task.sleep(interval: 1)
                 
                 guard let self = self,
                       gesture.state == .began || gesture.state == .changed else { return }
--- a/Adamant/Modules/Chat/View/Subviews/ChatBaseMessage/ChatMessageCell.swift
+++ b/Adamant/Modules/Chat/View/Subviews/ChatBaseMessage/ChatMessageCell.swift
@@ -499,7 +499,7 @@ private extension ChatMessageCell {
             messageContainerView.animatePressDown()
             
             Task { [weak self] in
-                try? await Task.sleep(interval: 1)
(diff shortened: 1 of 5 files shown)


Adamant/Modules/Chat/View/Subviews/ChatBaseMessage/ChatMessageCell.swift
```diff
@@ -499,7 +499,7 @@ private extension ChatMessageCell {
             messageContainerView.animatePressDown()
             
             Task { [weak self] in
-                try? await Task.sleep(nanoseconds: UInt64(1.5) * 1_000_000_000)
+                try? await Task.sleep(interval: 1)
```
adamantmm: Make a constant with a name and use it everywhere

adamantmm (changes requested):

Adamant/Modules/Chat/View/Subviews/ChatBaseMessage/ChatMessageCell.swift
vovameister: done

adamantmm (approved):

Lainaaa (approved):

PR #828, opened 2025-04-21
[trello.com/c/iGILackD] fix flow with remote reaction

vovameister:
fix flow with new reaction remote notification tap
also fix go to reply message highlight

### DIFF
--- a/Adamant/Modules/ChatsList/ChatListViewController.swift
+++ b/Adamant/Modules/ChatsList/ChatListViewController.swift
@@ -441,22 +441,33 @@ final class ChatListViewController: KeyboardObservingViewController {
             navigationController?.pushViewController(controller, animated: true)
         }
     }
-
+    
     // MARK: Helpers
     func chatViewController(
         for chatroom: Chatroom,
         with messageId: String? = nil,
         newChat: Bool = false,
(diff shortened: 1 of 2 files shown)


Adamant/Modules/ChatsList/ChatListViewController.swift
```diff
@@ -441,22 +441,33 @@ final class ChatListViewController: KeyboardObservingViewController {
...
         for chatroom: Chatroom,
         with messageId: String? = nil,
         newChat: Bool = false,
-        animateMessageType: MessageAnimationType = MessageAnimationType.none
+        animateMessageType: MessageAnimationType = .none
     ) -> ChatViewController {
         let vc = screensFactory.makeChat()
         vc.hidesBottomBarWhenPushed = true
+
+        var idAndAnimationType: (String?, MessageAnimationType) = (
```
Lainaaa, line 455: I don't like this style and prefer to use two different variables instead.

Lainaaa (approved):
Take a look at comment

Adamant/Modules/ChatsList/ChatListViewController.swift
vovameister: it's more convenient to perceive it this way, since the method messageId() returns it in exactly this form

PR #829, opened 2025-04-21
[trello.com/c/tVl3WiJp] fix popup while upload more chats

vovameister:
isInitiallySynced has notification that recall popup so now it has not extra didSet
also remove regular messages preloading (for 5 chats) after 1st load

### DIFF
--- a/Adamant/Services/DataProviders/AdamantChatsProvider.swift
+++ b/Adamant/Services/DataProviders/AdamantChatsProvider.swift
@@ -375,22 +375,29 @@ extension AdamantChatsProvider {
             privateKey: privateKey
         )
         
-        isInitiallySynced = true
-        preLoadChats(array, address: address)
+        if !isInitiallySynced {
+            isInitiallySynced = true
+            preLoadChats(array, address: address, withInitial: true)
+        } else {
+            preLoadChats(array, address: address)
+        }
 
         setState(.upToDate, previous: prevState)
     }
 
-    func preLoadChats(_ array: [Transaction], address: String) {
+    func preLoadChats(_ array: [Transaction], address: String, withInitial: Bool = false) {
         let prioritized = array.filter { transaction in
             removedMessages.contains(String(transaction.id))
         }
 
-        let regular = array.filter { transaction in
-            !removedMessages.contains(String(transaction.id))
-        }.prefix(preLoadChatsCount)
+        let regular: ArraySlice<Transaction> = withInitial
+            ? array.filter { transaction in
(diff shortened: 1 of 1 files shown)


adamantmm (dismissed):
> also remove regular messages preloading (for 5 chats) after 1st load

Why did you do this?

Adamant/Services/DataProviders/AdamantChatsProvider.swift
```diff
@@ -375,22 +375,29 @@ extension AdamantChatsProvider {
             privateKey: privateKey
         )
         
-        isInitiallySynced = true
-        preLoadChats(array, address: address)
+        if !isInitiallySynced {
```
Lainaaa, line 378: preLoadChats(array, address: address, withInitial: !isInitiallySynced)
            isInitiallySynced = true

Adamant/Services/DataProviders/AdamantChatsProvider.swift
```diff
@@ -375,22 +375,29 @@ extension AdamantChatsProvider {
...
+            preLoadChats(array, address: address, withInitial: true)
+        } else {
+            preLoadChats(array, address: address)
+        }
 
         setState(.upToDate, previous: prevState)
     }
 
-    func preLoadChats(_ array: [Transaction], address: String) {
+    func preLoadChats(_ array: [Transaction], address: String, withInitial: Bool = false) {
```
Lainaaa, line 388: If `isInitiallySynced` available like a class field why do you need it? 
just use `!isInitiallySynced` instead

Lainaaa (commented):
I am not sure about my comments but at least take a look and lets improve if possible

vovameister:
> > also remove regular messages preloading (for 5 chats) after 1st load
> 
> Why did you do this?

Previously, we used to load messages for 5 chats out of 20 only on the first download. I don't see the point in adding this for subsequent downloads, especially since I removed the loader in this PR for subsequent downloads, with a slow Internet this can lead to unpredictable consequences

Adamant/Services/DataProviders/AdamantChatsProvider.swift
vovameister: This is more confusing than it is improving

Adamant/Services/DataProviders/AdamantChatsProvider.swift
vovameister: the change in isInitiallySynced value happens at the wrong moment, its not gonna work right

Lainaaa (approved):

PR #830, opened 2025-04-21
[trello.com/c/JnQ8ArFZ] add update last transaction in mapper

vovameister:
add update chatroom last transaction after read reaction for hidden message to insta update chatList


Lainaaa (commented):
Are you sure about this approach?
It seems that CoreDataRealationMapper should make this work
We already do it in AdamantChatsProvider.swift:2081
Maybe it is possible to make this in a more appropriate place?

Lainaaa (approved):

PR #834, opened 2025-04-24
[trello.com/c/dEf1yVQr] Allow sending the first message faster

Lainaaa:
* Use intervals from wallets-script instead of hardcode. 
* Improve overall logic.
* Remove `forceUpdateAllBalances` and `walletUpdatedNotification`.

### DIFF
--- a/Adamant/Modules/Wallets/Bitcoin/BtcWalletService.swift
+++ b/Adamant/Modules/Wallets/Bitcoin/BtcWalletService.swift
@@ -157,7 +157,6 @@ final class BtcWalletService: WalletCoreProtocol, WalletStaticCoreProtocol, @unc
     private let walletPath = "m/44'/0'/21'/0/0"
 
     // MARK: - Notifications
-    let walletUpdatedNotification = Notification.Name("adamant.btcWallet.walletUpdated")
     let serviceEnabledChanged = Notification.Name("adamant.btcWallet.enabledChanged")
     let serviceStateChanged = Notification.Name("adamant.btcWallet.stateChanged")
     let transactionFeeUpdated = Notification.Name("adamant.btcWallet.feeUpdated")
@@ -317,12 +316,6 @@ final class BtcWalletService: WalletCoreProtocol, WalletStaticCoreProtocol, @unc
         
         walletUpdateSender.send()
         
-        NotificationCenter.default.post(
-            name: walletUpdatedNotification,
-            object: self,
-            userInfo: [AdamantUserInfoKey.WalletService.wallet: wallet]
-        )
-        
         setState(.upToDate)
 
         if let rate = try? await getFeeRate() {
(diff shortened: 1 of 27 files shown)


joao-krauchenco-mottu (approved):

vovameister (approved):

Adamant/Modules/Wallets/Bitcoin/BtcWalletService.swift
```diff
@@ -386,8 +394,8 @@ final class BtcWalletService: WalletCoreProtocol, WalletStaticCoreProtocol, @unc
         wallet.isBalanceInitialized = true
 
         balanceInvalidationSubscription = Task { [weak self] in
-            try await Task.sleep(interval: Self.balanceLifetime, pauseInBackground: true)
-            guard let self else { return }
+            guard let self = self else { return }
+            try await Task.sleep(interval: Double(self.balanceValidInterval ?? Self.balanceLifetime) / 1000, pauseInBackground: true)
```
adamantmm: Looks like `self.balanceValidInterval ?? Self.balanceLifetime` is excessive.
Can we delete `Self.balanceLifetime` fully? As it's hardcoded to `static var balanceLifetime: Int { 300000 } // 5 minutes` and do the same thing?

adamantmm (changes requested):

Adamant/Modules/Wallets/Bitcoin/BtcWalletService.swift
adamantmm: If you use `Self.balanceLifetime` as a fallback when `self.balanceValidInterval` is not set, rename `Self.balanceLifetime` to `self.balanceValidIntervalDefault`.

Adamant/Services/WalletAutoUpdaterService.swift
```diff
@@ -133,41 +133,11 @@ actor WalletAutoUpdateService {
...
-                case .erc20, .dash:     return 25
-                case .doge:             return 30
-                case .btc, .kly:        return 50
-                case .admNewAccount:    return 2
+        if let isNewAccount = accountService.account?.isNewAccount, isNewAccount {
+            if let wallet = wallet as? AdmWalletService {
+                return Double(wallet.balanceCheckIntervalNewAccount ?? 50000) / 1000
             }
         }
+        return Double(wallet.balanceCheckInterval ?? 50000) / 1000
```
adamantmm, line 144: 50000 ⟶ `wallet.balanceCheckIntervalDefault`.
No magic numbers.

Adamant/Modules/Wallets/Bitcoin/BtcWalletService.swift
Lainaaa: > If you use Self.balanceLifetime as a fallback when self.balanceValidInterval is not set, rename Self.balanceLifetime to self.balanceValidIntervalDefault.

Yes, this is the idea

I will, `balanceLifetime` was used before as a naming

Adamant/Services/WalletAutoUpdaterService.swift
```diff
@@ -109,6 +109,9 @@ actor WalletAutoUpdateService {
                 interval: getTimeIntervalFor(wallet: core),
                 queue: .global(qos: .utility),
                 callback: {
+                    if wallet.core is AdmWalletService{
+                        self.accountService.update()
```
Lainaaa, line 113: Because `accountService` is responsible for updating adm instead of `AdmWalletService` in the current implementation

PR #840, opened 2025-04-28
[trello.com/c/7YVA2NTe] update cancel file uploading

vovameister:
Now the entire loading of a message with a file is canceled by one click on the cross next to the message. Animation of the general progress of file loading has been added.

### DIFF
--- a/Adamant/Modules/Chat/View/Helpers/FileMessageStatus.swift
+++ b/Adamant/Modules/Chat/View/Helpers/FileMessageStatus.swift
@@ -17,7 +17,7 @@ enum FileMessageStatus: Equatable {
 
     var image: UIImage {
         switch self {
-        case .busy: return .asset(named: "status_pending") ?? .init()
+        case .busy: return .asset(named: "status_failed") ?? .init()
         case .success: return .asset(named: "status_success") ?? .init()
         case .failed: return .asset(named: "status_failed") ?? .init()
         case let .needToDownload(failed):
--- a/Adamant/Modules/Chat/View/Subviews/ChatMedia/Container/ChatMediaContainerView.swift
+++ b/Adamant/Modules/Chat/View/Subviews/ChatMedia/Container/ChatMediaContainerView.swift
@@ -9,6 +9,7 @@
 import Combine
 import CommonKit
 import UIKit
+import SwiftUI
 
 final class ChatMediaContainerView: UIView {
     private let spacingView: UIView = {
@@ -88,9 +89,28 @@ final class ChatMediaContainerView: UIView {
         button.addTarget(self, action: #selector(onStatusButtonTap), for: .touchUpInside)
         return button
     }()
+    
+    private lazy var progressRingHostingView: UIView = {
+        let progressBar = CircularProgressView { [weak self] in
+            guard let self else { return .init(progress: .zero, hidden: true) }
+            return self.statusProgressState
+        }
+        let hosting = UIHostingController(rootView: progressBar)
+        hosting.view.backgroundColor = .clear
+        hosting.view.isUserInteractionEnabled = false
+        return hosting.view
+    }()
 
     private lazy var contentView = ChatMediaContentView()
     private lazy var chatMenuManager = ChatMenuManager(delegate: self)
+    
+    private var statusProgressState = CircularProgressState(
+        lineWidth: 2.0,
+        backgroundColor: .clear,
+        progressColor: .adamant.primary,
+        progress: .zero,
+        hidden: true
+    )
 
     // MARK: Dependencies
 
@@ -123,6 +143,11 @@ final class ChatMediaContainerView: UIView {
     }
 
     @objc func onStatusButtonTap() {
+        if model.status == .busy {
+            actionHandler(.cancelUploading(messageId: model.id))
+            return
+        }
+        
         if model.status == .failed,
             let file = model.content.fileModel.files.first
         {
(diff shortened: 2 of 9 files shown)


Adamant/Modules/Chat/View/Helpers/FileMessageStatus.swift
```diff
@@ -17,7 +17,7 @@ enum FileMessageStatus: Equatable {
 
     var image: UIImage {
         switch self {
-        case .busy: return .asset(named: "status_pending") ?? .init()
+        case .busy: return .asset(named: "status_failed") ?? .init()
```
adamantmm, line 20: Isn't it misleading?

Adamant/Modules/Chat/View/Subviews/ChatMedia/Container/ChatMediaContainerView.swift
```diff
@@ -88,9 +89,28 @@ final class ChatMediaContainerView: UIView {
         button.addTarget(self, action: #selector(onStatusButtonTap), for: .touchUpInside)
         return button
     }()
+    
+    private lazy var progressRingHostingView: UIView = {
+        let progressBar = CircularProgressView { [weak self] in
+            guard let self else { return .init(progress: .zero, hidden: true) }
+            return self.statusProgressState
+        }
+        let hosting = UIHostingController(rootView: progressBar)
```
Lainaaa, line 98: I don't really like an approach to mix swiftui with existed UIKit code while we already use 3 approaches:
- SnapKit
- UIKit programmatically
- Storyboards

Lets use one of the listed above approaches instead

Adamant/Modules/Chat/View/Subviews/ChatMedia/Container/ChatMediaContainerView.swift
```diff
@@ -166,13 +197,35 @@ extension ChatMediaContainerView {
         updateOwnReaction()
         updateOpponentReaction()
         updateStatus(model.status)
+        updateProgressRing()
```
Lainaaa, line 200: I like your approach but may we call it only in case where message contains files instead of making it always?

Adamant/Modules/Chat/ViewModel/ChatFileService.swift
```diff
@@ -204,17 +204,24 @@ final class ChatFileService: ChatFileProtocol, Sendable {
...
-            )
+            fileIdsToRemove.append(fileId)
         }
+
+        await removeFilesFromRichFile(
+            oldIds: fileIdsToRemove,
+            txId: messageId
+        )
+
+        uploadTasks[messageId] = nil
```
Lainaaa, line 225: Did you check that task was cancelled? Because of the first view it seems that instead of cancelling task it just makes `nil` the value of the dictionary

Adamant/Modules/Chat/ViewModel/ChatFileService.swift
```diff
@@ -49,7 +49,7 @@ final class ChatFileService: ChatFileProtocol, Sendable {
     private var fileDownloadAttemptsCount: [String: Int] = [:]
     private var uploadingFilesDictionary: [String: FileMessage] = [:]
     private var previewDownloadsAttemps: [String: Int] = [:]
-    private var uploadTasks: [String: Task<UploadFileResult, Error>] = [:]
+    private var uploadTasks: [String: [String: Task<UploadFileResult, Error>]] = [:]
```
Lainaaa: Did you consider to use taskManager?

Lainaaa: Lets make a comment explaining this complex data structure

Lainaaa (commented):
Overall is good but could be improved as I think

Adamant/Modules/Chat/View/Helpers/FileMessageStatus.swift
vovameister: "status_failed" is just a name of the picture in our project, it is used in many places, I can try to rename it manually to like "cross in a circle" or just duplicate it with new name for this view. but to be honest I think it's just the title of the picture it just help to find it in Assets

Adamant/Modules/Chat/View/Subviews/ChatMedia/Container/ChatMediaContainerView.swift
vovameister: this type of messages always contains files

Adamant/Modules/Chat/ViewModel/ChatFileService.swift
vovameister: task.cancel() string 213

Adamant/Modules/Chat/View/Subviews/ChatMedia/Container/ChatMediaContainerView.swift
vovameister: we already have this view in this cell (MediaContentView/FileListContentView) I just wanna reuse CircularProgressView. why should i write it on UIkit if its working fine on swift UI and used before this task

Adamant/Modules/Chat/ViewModel/ChatFileService.swift
vovameister: its [messageId: [fileID: Task]] to cancel all tasks by messageId 
the upload itself is managed by fileID

Lainaaa: Yes, lets add this explanation

Lainaaa (dismissed):

art-divin (approved):

PR #841, opened 2025-04-29
[trello.com/c/sG80IZan] fix reading while TransactionDetails on screen

vovameister:
now if we have TransactionDetails or TransferViewController on top оf the chat, messages in chat will not be available to read and reeding will be triggered after dissmiss

### DIFF
--- a/Adamant/Modules/Chat/View/ChatViewControllerState.swift
+++ b/Adamant/Modules/Chat/View/ChatViewControllerState.swift
@@ -20,12 +20,17 @@ struct ChatViewControllerState {
     var isAppActive = true
     var isScrollingToBottom = false
     var shouldScrollToNewMessagesOrSavedPosition = true
+    var presentedControllerIDs: Set<String> = [] {
+        didSet {
+            print("can read:", presentedControllerIDs.isEmpty)
+        }
+    }
     
     //calculation for animation, might use for something else in the future
     var isAnimationAllowed: Bool {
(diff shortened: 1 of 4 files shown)


Adamant/Modules/Chat/View/ChatViewControllerState.swift
```diff
@@ -20,12 +20,17 @@ struct ChatViewControllerState {
     var isAppActive = true
     var isScrollingToBottom = false
     var shouldScrollToNewMessagesOrSavedPosition = true
+    var presentedControllerIDs: Set<String> = [] {
```
Lainaaa:
```suggestion
    var presentedControllerIDs: Set<String> = []
```

Lainaaa (commented):
I think it should work, however this code will be harder to maintenance because devs have to keep in mind to update this code when they would want to add new screen above the chat. 

Instead, what about just checking if the current screen is on the top of the navigation stack?

vovameister:
> navigation stack

its now working on MacOS because of split view

Adamant/Modules/Chat/View/ChatViewControllerState.swift
vovameister: print deleted

Lainaaa:
> > navigation stack
> 
> its now working on MacOS because of split view

Should exist something else for MacOS I think
Hope [this](https://arc.net/l/quote/imbfsanj) can help

PR #846, opened 2025-05-05
[trello.com/c/aZtAfZ44] fix double remote noti after read

vovameister:
after reading a message, the remote notification for this message will not be displayed


Adamant/Models/CoreData/Chatroom+CoreDataClass.swift
```diff
@@ -9,16 +9,20 @@
...
 
     func markAsReaded() {
         hasUnreadMessages = false
-
+        
         if let trs = transactions as? Set<ChatTransaction> {
-            trs.filter { $0.isUnread }.forEach { $0.isUnread = false }
+            trs.filter { $0.isUnread }.forEach {
+                $0.isUnread = false
+                UserDefaultsManager.addLastReadId($0.transactionId)
```
Lainaaa: Why do you want to add each separately instead of adding all of them in one time?

vovameister: because i only need to follow the messages i just read and add it to userDefoult, not every single message

Adamant/Models/CoreData/Chatroom+CoreDataClass.swift
```diff
@@ -28,9 +30,16 @@ public class Chatroom: NSManagedObject, @unchecked Sendable {
         else {
             return
         }
-        message.isUnread = false
+        
+        if message.isUnread {
```
Lainaaa:
```suggestion
    message.isUnread = false
```

Adamant/Models/CoreData/Chatroom+CoreDataClass.swift
```diff
@@ -28,9 +30,16 @@ public class Chatroom: NSManagedObject, @unchecked Sendable {
...
+            message.isUnread = false
+        }
 
-        message.richMessageTransactions?.forEach { $0.isUnread = false }
+        message.richMessageTransactions?.forEach {
+            if $0.isUnread {
```
Lainaaa: Why do you need this?
return it back

Adamant/Models/CoreData/Chatroom+CoreDataClass.swift
Lainaaa: Why do you need this?
return it back

CommonKit/Sources/CommonKit/Core/UserDefaultsManager.swift
```diff
@@ -0,0 +1,29 @@
...
+//  CommonKit
+//
+//  Created by Владимир Клевцов on 2. 5. 2025..
+//
+public struct UserDefaultsManager {
+    @UserDefaultsStorage(.needsToShowNoActiveNodesAlert)
+    static var needsToShowNoActiveNodesAlert: Bool?
+
+    @UserDefaultsStorage(.lastReceivedId)
+    public static var lastReceivedId: [String]?
```
Lainaaa, line 12: Why do you use array instead of `Set`?

Adamant/Services/DataProviders/AdamantChatsProvider.swift
```diff
@@ -1922,6 +1923,9 @@ extension AdamantChatsProvider {
             if unreadTransactions.count == 0 {
                 unreadTransactions = newMessageTransactions.filter { $0.height == 0 }
             }
+            unreadTransactions.forEach {
```
Lainaaa, line 1926: What is your approach? Save every unread message to a local storage  and after check if the received message contains this id - cool! But what about deleting the previous saved messages? 
Also, why do you need to make additional savings if every message is saving to a Core Data?

Lainaaa (changes requested):
Check comments

CommonKit/Sources/CommonKit/Core/UserDefaultsManager.swift
vovameister: userDefault cant store Set-_-

Adamant/Services/DataProviders/AdamantChatsProvider.swift
vovameister: because in the background I don't have such access to core data? and the id will start to erase itself when the limit of 100 is exceeded, look at the code more carefully

CommonKit/Sources/CommonKit/Core/UserDefaultsManager.swift
Lainaaa: Use <Set> everywhere it possible but when work with UserDefaults just convert to Array?

Lainaaa (commented):
Describe what you did and how
I spend too much time for this myself while it much easier just to make a secription

Lainaaa (approved):
lgtm

Lainaaa (approved):
lgtm

PR #850, opened 2025-05-07
[trello.com/c/7YVA2NTe] cancel files loading update

vovameister:
1. fix index replace when sending fail (resend message bug)
2. update animation for uploading
3. remove animation and bring back icon when file downloaind
4. fix bug with part file send when user has a slow internet connection

### DIFF
--- a/Adamant/Modules/Chat/View/Helpers/CircularProgressView.swift
+++ b/Adamant/Modules/Chat/View/Helpers/CircularProgressView.swift
@@ -8,6 +8,7 @@
 
 import SwiftUI
 import UIKit
+import Combine
 
 final class CircularProgressState: ObservableObject {
     let lineWidth: CGFloat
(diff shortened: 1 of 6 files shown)


Adamant/Modules/Chat/View/Helpers/CircularProgressView.swift
```diff
@@ -33,18 +36,31 @@ final class CircularProgressState: ObservableObject {
...
-
+    
     var body: some View {
         ZStack {
-            Circle()
-                .stroke(
-                    Color(uiColor: state.backgroundColor),
-                    lineWidth: state.lineWidth
-                )
+            if let gradient = state.backgroundGradient {
```
Lainaaa, line 49: Use ternary operator if possible

Lainaaa (approved):

PR #852, opened 2025-05-08
[trello.com/c/jKqpxiCa-5] fix macOc hidden height for message reading

vovameister:
also increased new message line thickness

### DIFF
--- a/Adamant/Modules/Chat/View/Subviews/NewMessagesCell.swift
+++ b/Adamant/Modules/Chat/View/Subviews/NewMessagesCell.swift
@@ -34,7 +34,7 @@ final class NewMessagesCell: MessageReusableView {
         separatorLine.translatesAutoresizingMaskIntoConstraints = false
         label.translatesAutoresizingMaskIntoConstraints = false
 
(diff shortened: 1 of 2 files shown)


Lainaaa (changes requested):
typo `hiden` -> `hidden`

Adamant/Modules/Chat/View/Subviews/NewMessagesCell.swift
```diff
@@ -34,7 +34,7 @@ final class NewMessagesCell: MessageReusableView {
         separatorLine.translatesAutoresizingMaskIntoConstraints = false
         label.translatesAutoresizingMaskIntoConstraints = false
 
-        let pixelSize = 1 / UIScreen.main.scale
+        let pixelSize = 2 / UIScreen.main.scale
```
Lainaaa, line 37: Should be good, however, I don't fully understand this line, coulde you explain please

Lainaaa: Ideally would be to put comment here

Lainaaa (approved):

PR #853, opened 2025-05-08
[trello.com/c/jKqpxiCa] update save position

vovameister:
return save position by Y, but if we have a new messages in chat it will will recalculate the difference
its not working same way an IOS and Mac
ios working well and save position perfectly with new messages or without.
MasOs work fine only without new messages, because it cant restore content height as ios due to rendering features


Lainaaa:
Screen Recording 2025-05-08 at 4.08.35 PM.mov

Does not work as expected, details you can find on the screen recording

Lainaaa (changes requested):
Does not work as expected, details you can find on the trello card

Lainaaa (approved):

PR #855, opened 2025-05-12
[trello.com/c/1lX3h2xx] fix getErc20 status and timestamp

vovameister:
fix getErc20 throw if we dont take receipt or details details (both)
add post request for block.timestamp

### DIFF
--- a/Adamant/Modules/Wallets/ERC20/ERC20WalletService.swift
+++ b/Adamant/Modules/Wallets/ERC20/ERC20WalletService.swift
@@ -452,78 +452,67 @@ extension ERC20WalletService: SwinjectDependentService {
 extension ERC20WalletService {
     func getTransaction(by hash: String, waitsForConnectivity: Bool) async throws -> EthTransaction {
         let sender = wallet?.address
-        let isOutgoing: Bool
-
-        // MARK: 1. Transaction details
-        let details: Web3Core.TransactionDetails = try await erc20ApiService.requestWeb3(
-            waitsForConnectivity: waitsForConnectivity
-        ) { web3 in
+        var details: Web3Core.TransactionDetails?
+        var receipt: TransactionReceipt?
+        
+        details = try await erc20ApiService.requestWeb3(waitsForConnectivity: waitsForConnectivity) { web3 in
             try await web3.eth.transactionDetails(hash)
         }.get()
-
-        let receipt = try await erc20ApiService.requestWeb3(
-            waitsForConnectivity: waitsForConnectivity
-        ) { web3 in
+        
+        receipt = try await erc20ApiService.requestWeb3(waitsForConnectivity: waitsForConnectivity) { web3 in
             try await web3.eth.transactionReceipt(hash)
         }.get()
-
-        // MARK: 3. Check if transaction is delivered
-        guard receipt.status == .ok,
-            let blockNumber = details.blockNumber
-        else {
-            let transaction = details.transaction.asEthTransaction(
-                date: nil,
-                gasUsed: receipt.gasUsed,
-                gasPrice: receipt.effectiveGasPrice,
-                blockNumber: nil,
-                confirmations: nil,
-                receiptStatus: receipt.status,
-                isOutgoing: false
-            )
-            return transaction
+        
+        guard details != nil || receipt != nil else {
+            throw WalletServiceError.remoteServiceError(message: "Both transaction details and receipt are unavailable.")
         }
 
-        // MARK: 4. Block timestamp & confirmations
-        let currentBlock = try await erc20ApiService.requestWeb3(
-            waitsForConnectivity: waitsForConnectivity
-        ) { web3 in
-            try await web3.eth.blockNumber()
-        }.get()
+        let isOutgoing: Bool = {
+            guard let sender = sender else { return false }
+            return details?.transaction.sender?.address == sender
+        }()
 
-        let block = try await erc20ApiService.requestWeb3(
-            waitsForConnectivity: waitsForConnectivity
-        ) { web3 in
-            try await web3.eth.block(by: receipt.blockHash)
-        }.get()
+        let tx = details?.transaction
+        let timestamp: Date? = await {
+            guard let blockNumber = receipt?.blockNumber else { return nil }
+            let blockHex = "0x" + String(blockNumber, radix: 16)
+            do {
+                return try await fetchBlockTimestamp(blockNumberHex: blockHex)
+            } catch {
+                print("❌ Failed to fetch block timestamp:", error)
+                return nil
+            }
+        }()
 
-        guard currentBlock >= blockNumber else {
-            throw WalletServiceError.remoteServiceError(
-                message: "ERC20 confirmations calculating error"
-            )
-        }
+        let confirmations: String? = await {
+            guard let blockNumber = details?.blockNumber else { return nil }
+            do {
+                let currentBlock = try await erc20ApiService.requestWeb3(waitsForConnectivity: waitsForConnectivity) { web3 in
+                    try await web3.eth.blockNumber()
+                }.get()
+                return String(currentBlock - blockNumber)
+            } catch {
+                return nil
+            }
+        }()
 
-        let confirmations = currentBlock - blockNumber
-
-        let transaction = details.transaction
-
-        if let sender = sender {
-            isOutgoing = transaction.sender?.address == sender
-        } else {
-            isOutgoing = false
-        }
-
-        let ethTransaction = transaction.asEthTransaction(
-            date: block.timestamp,
-            gasUsed: receipt.gasUsed,
-            gasPrice: receipt.effectiveGasPrice,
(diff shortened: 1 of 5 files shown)


Lainaaa:
There is no such a card in Trello

Lainaaa:
Is PR ready fir review?

Lainaaa:
Instead of commiting something like `delete print` lets just amend it and force push it

Adamant/Modules/Wallets/Ethereum/EthWalletService.swift
```diff
@@ -683,15 +683,20 @@ extension EthWalletService {
             let currentBlock = try await ethApiService.requestWeb3(waitsForConnectivity: false) { web3 in
                 try await web3.eth.blockNumber()
             }.get()
-
-            let block = try await ethApiService.requestWeb3(waitsForConnectivity: false) { web3 in
-                try await web3.eth.block(by: receipt.blockHash)
-            }.get()
+            
+            let timestamp: Date? = await {
+                let blockHex = "0x" + String(receipt.blockNumber, radix: 16)
```
Lainaaa, line 688: Could we incapsulate this logic and send just `receipt.blockNumber` to the `fetchBlockTimestamp(blockNumber: String)`

Adamant/Modules/Wallets/Ethereum/EthWalletService.swift
```diff
@@ -830,6 +836,38 @@ extension EthWalletService {
     func updateStatus(for id: String, status: TransactionStatus?) {
         coinStorage.updateStatus(for: id, status: status)
     }
+    
+    func fetchBlockTimestamp(blockNumberHex: String) async throws -> Date {
+        let body: [String: Any] = [
```
Lainaaa: Lets make it how it is currently in our code without new approaches like this

I mean make request by analogy as previous requests

Adamant/Modules/Wallets/Ethereum/EthWalletService.swift
```diff
@@ -830,6 +836,38 @@ extension EthWalletService {
...
+                origin: origin,
+                path: "",
+                method: .post,
+                jsonParameters: body,
+                timeout: .common
+            )
+            return .success(result)
+        }.get()
+
+        guard
```
Lainaaa: Same here, I don't think that `ethWalletService` should be responsible for decoding the data

Adamant/Modules/Wallets/ERC20/ERC20WalletService.swift
```diff
@@ -667,4 +656,42 @@ extension ERC20WalletService {
...
+        else {
+            throw WalletServiceError.remoteServiceError(message: "Invalid timestamp in block response")
+        }
+
+        return Date(timeIntervalSince1970: TimeInterval(timestampInt))
+    }
+}
+
+private extension String {
+    func stripHexPrefix() -> String {
```
Lainaaa: What does it mean `strip`? Are you sure that this is should be here? It seems that it is more data extension neither string in erc20walletservice

Adamant/Modules/Wallets/ERC20/ERC20WalletService.swift
```diff
@@ -667,4 +656,42 @@ extension ERC20WalletService {
     func updateStatus(for id: String, status: TransactionStatus?) {
         coinStorage.updateStatus(for: id, status: status)
     }
+    
+    func fetchBlockTimestamp(blockNumberHex: String) async throws -> Date {
```
Lainaaa: Same as for `EthWalletService`

Adamant/Modules/Wallets/ERC20/ERC20WalletService.swift
```diff
@@ -452,78 +452,66 @@ extension ERC20WalletService: SwinjectDependentService {
 extension ERC20WalletService {
     func getTransaction(by hash: String, waitsForConnectivity: Bool) async throws -> EthTransaction {
         let sender = wallet?.address
-        let isOutgoing: Bool
-
-        // MARK: 1. Transaction details
-        let details: Web3Core.TransactionDetails = try await erc20ApiService.requestWeb3(
-            waitsForConnectivity: waitsForConnectivity
-        ) { web3 in
+        var details: Web3Core.TransactionDetails?
```
Lainaaa:
```suggestion
        var details: Web3Core.TransactionDetails? = try await erc20ApiService.requestWeb3(waitsForConnectivity: waitsForConnectivity) { web3 in
            try await web3.eth.transactionDetails(hash)
        }.get()

        var receipt: TransactionReceipt? = try await erc20ApiService.requestWeb3(waitsForConnectivity: waitsForConnectivity) { web3 in
```

Adamant/Modules/Wallets/ERC20/ERC20WalletService.swift
```diff
@@ -452,78 +452,66 @@ extension ERC20WalletService: SwinjectDependentService {
...
-            receiptStatus: receipt.status,
+        let isOutgoing: Bool = {
+            guard let sender = sender else { return false }
+            return details?.transaction.sender?.address == sender
+        }()
+
+        let tx = details?.transaction
+        let timestamp: Date? = await {
+            guard let blockNumber = receipt?.blockNumber else { return nil }
+            let blockHex = "0x" + String(blockNumber, radix: 16)
```
Lainaaa, line 474: same as below

Lainaaa (commented):
Let’s strive to maintain the code cleanliness and the practices already in use in the project, and if you want to introduce something, let’s discuss it beforehand.

Lainaaa:
Overall it works, but the code have to be improved

Adamant/Modules/Wallets/Ethereum/EthWalletService.swift
vovameister: look at getTransactionHistory,  its the same besides decoding, I just dont want to write decoding stuct just for one field 
I think that the code I wrote is easier to log and debug than requests that call several abstract protocols, here it is unnecessary

Lainaaa: It is necessary because we already do like this, using different approaches on different parts of the app specially in one VC - that as really hard.

Also, when I search "getTransactionHistory" in the workspace it does not show me any results.

Lainaaa (changes requested):

Adamant/Modules/Wallets/ERC20/ERC20WalletService.swift
```diff
@@ -667,4 +653,37 @@ extension ERC20WalletService {
...
+            throw WalletServiceError.remoteServiceError(message: "Invalid timestamp in block response")
+        }
+
+        return Date(timeIntervalSince1970: TimeInterval(timestampInt))
+    }
+}
+
+extension String {
+    func trimHexPrefix() -> String {
+        hasPrefix("0x") ? String(dropFirst(2)) : self
```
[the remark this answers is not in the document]
art-divin: Agree with the proposal, better to streamline the logic in one place regarding similar functionality, i.e. timestamps

Adamant/Modules/Wallets/Ethereum/EthWalletService.swift
```diff
@@ -830,6 +836,33 @@ extension EthWalletService {
     func updateStatus(for id: String, status: TransactionStatus?) {
         coinStorage.updateStatus(for: id, status: status)
     }
+    
+    func fetchBlockTimestamp(blockNumberHex: String) async throws -> Date {
```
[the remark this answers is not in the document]
art-divin: Is it possible (reasonable) to move fetchBlockTimestamp into the superclass?
or maybe a protocol extension for both these services. 

Just thinking out loud, maybe you tried it already 🤷🏻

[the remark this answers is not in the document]
vovameister: moved it to EthApiService, not sure if it looks great in this abstraction, but not so bad in extention

Adamant/Modules/Wallets/ERC20/ERC20WalletService.swift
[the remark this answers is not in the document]
vovameister: resolved

Adamant/Modules/Wallets/ERC20/ERC20WalletService.swift
```diff
@@ -452,78 +452,63 @@ extension ERC20WalletService: SwinjectDependentService {
...
-        let confirmations = currentBlock - blockNumber
-
-        let transaction = details.transaction
-
-        if let sender = sender {
-            isOutgoing = transaction.sender?.address == sender
-        } else {
-            isOutgoing = false
+        
+        guard details != nil || receipt != nil else {
```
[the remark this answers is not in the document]
vovameister: it seems like without details?.transaction there is no way to execute the method with the current implementation of asEthTransaction so it doesn't make sense

Lainaaa (approved):

PR #860, opened 2025-05-14
[trello.com/c/jKqpxiCa/12] chat scroll and reading update

vovameister:
review by commits

### DIFF
--- a/Adamant/Modules/Chat/View/ChatViewController.swift
+++ b/Adamant/Modules/Chat/View/ChatViewController.swift
@@ -811,7 +811,6 @@ extension ChatViewController {
                 viewModel.animationType = MessageAnimationType.none
                 let isFirstMessagesInChat = viewModel.unreadMessagesIds?.count == viewModel.messages.count
                 scrollToPosition(.messageId(unreadMessage), setExtraOffset: !isFirstMessagesInChat, scrollAt: .top)
-
             }
         }
     }
@@ -881,6 +880,22 @@ extension ChatViewController {
             break
         }
     }
+    
+    fileprivate func markMessageFromCurrentToBottomAsRead() {
+        guard let unreadIndexes = viewModel.unreadMesaggesIndexes, !unreadIndexes.isEmpty else { return }
+
+        let visibleSections = messagesCollectionView.indexPathsForVisibleItems.map { $0.section }
+        guard let minSection = visibleSections.min() else { return }
+
+        let totalSections = messagesCollectionView.numberOfSections
+        guard totalSections > 0 else { return }
+
+        for section in minSection..<totalSections {
+            if unreadIndexes.contains(section) {
+                viewModel.markMessageAsRead(index: section)
+            }
+        }
+    }
 }
 
 // MARK: Making entities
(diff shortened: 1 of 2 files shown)


Adamant/Modules/Chat/View/ChatViewController.swift
```diff
@@ -220,7 +219,9 @@ final class ChatViewController: MessagesViewController {
 
     override func scrollViewDidScroll(_ scrollView: UIScrollView) {
         super.scrollViewDidScroll(scrollView)
-        updateUnreadMessages()
+        if state.canReadByDidScroll {
```
Lainaaa: What does it mean `canReadByDidScroll`?

Adamant/Modules/Chat/View/ChatViewController.swift
```diff
@@ -879,6 +883,22 @@ extension ChatViewController {
             break
         }
     }
+    
+    fileprivate func markMessageFromCurrentToBottomAsRead() {
```
Lainaaa: markAllUnreadMessagesAsRead?

Lainaaa: you don't need to describe UI here, you need to describe logic

Lainaaa: What do you think?

Adamant/Modules/Chat/View/ChatViewController.swift
vovameister: it means can we read message by scrollViewDidScroll() method

Adamant/Modules/Chat/View/ChatViewController.swift
vovameister: markAllUnreadMessagesAsRead? not all

vovameister: if we have under messages higher we will not read them

Adamant/Modules/Chat/View/ChatViewController.swift
Lainaaa: Still don't understand 👉👈

Adamant/Modules/Chat/View/ChatViewController.swift
Lainaaa: How could we read something below and don't read above?

Adamant/Modules/Chat/View/ChatViewControllerState.swift
```diff
@@ -20,6 +20,7 @@ struct ChatViewControllerState {
     var isAppActive = true
     var isScrollingToBottom = false
     var shouldScrollToNewMessages = true
+    var canReadByDidScroll = false
```
Lainaaa:
```suggestion
    var isInitialMessagesWereUpdated = false
```

Adamant/Modules/Chat/View/ChatViewController.swift
Lainaaa:
```suggestion
    fileprivate func markMessagesFromCurrentToBottomAsRead() {
```

Lainaaa (approved):

PR #863, opened 2025-05-19
[trello.com/c/7YVA2NTe] cancel uploading update

vovameister:
- fix progressRing macos layout
- progress ring animation show total upload volume instead of avarage volume

### DIFF
--- a/Adamant/Modules/Chat/View/Subviews/ChatMedia/Container/ChatMediaContainerView.swift
+++ b/Adamant/Modules/Chat/View/Subviews/ChatMedia/Container/ChatMediaContainerView.swift
@@ -214,9 +214,7 @@ extension ChatMediaContainerView {
     func totalUploadProgress() -> Double {
         let files = model.content.fileModel.files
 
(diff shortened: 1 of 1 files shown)


Adamant/Modules/Chat/View/Subviews/ChatMedia/Container/ChatMediaContainerView.swift
```diff
@@ -206,29 +211,35 @@ extension ChatMediaContainerView {
...
-    func averageUploadProgress() -> Double {
+    func totalUploadProgress() -> Double {
         let files = model.content.fileModel.files
-        
-        let progresses = files
-            .compactMap { $0.progress }
-        
-        guard !progresses.isEmpty else {
+
+        let totalBytes: Int64 = files.reduce(0) { result, file in
```
Lainaaa: Tip:
let totalBytes = files.map(\.file.size).reduce(0, +)

Lainaaa (approved):
If you can fast rewrite it saving previous progress instead of recalculating it every time it would be nice
Current implementation is okay, however performance could be improved

PR #864, opened 2025-05-19
[trello.com/c/aZtAfZ44] fix replace message

vovameister:
fix replace message and make read reaction faster


Lainaaa:
Explain you PR please

PR #868, opened 2025-05-23
[trello.com/c/aZtAfZ44] fix select row when we enter the chat via popup

vovameister:


Lainaaa (changes requested):
Could we move `updateSelectedRow` into the presenter and just on `presentChatroom` also call `updateSelectedRow`. We can make it via delegate pattern for example. Because now dev will need to know about your know method and call it when it want to add new functionality for example, lets care about future us and future other devs of this project!

Lainaaa (changes requested):
Thanks! The code look to be good!

However, make sure to name your commit! Lets care about our commit history.
So in this case I suggest to squash the commits.

art-divin (approved):

PR #872, opened 2025-05-27
[trello.com/c/sG80IZan] delay badge update for current chat

vovameister:


Adamant/Modules/ChatsList/ChatListViewController.swift
```diff
@@ -363,6 +368,16 @@ final class ChatListViewController: KeyboardObservingViewController {
                 self?.configureCell(cell, for: chatroom)
             }
             .store(in: &subscriptions)
+        
+        unreadBadgeUpdateSubject
+            .debounce(for: .seconds(currentChatBudgeUpdateDelay), scheduler: DispatchQueue.main)
+            .sink { [weak self] count in
+                guard let self = self else { return }
+                if let count {
```
Lainaaa: tip:
you could check count on the guard statement

Adamant/Modules/ChatsList/ChatListViewController.swift
```diff
@@ -140,6 +141,10 @@ final class ChatListViewController: KeyboardObservingViewController {
     private var loadNewChatTask: Task<(), Never>?
     private var subscriptions = Set<AnyCancellable>()
     private var swipedIndex: IndexPath?
+    
+    // MARK: Publishers
+    private var unreadBadgeUpdateSubject = ObservableSender<Int?>()
```
Lainaaa, line 146: tip:
In mark you write `Publishers` and just below you write `Subject`, i think it is better just to make it `unreadBadgeUpdatePublisher`

Lainaaa (commented):
Should not it be easier just to make Task.sleep instead of this hard logic?

Adamant/Modules/ChatsList/ChatListViewController.swift
vovameister: resolve

vovameister:
> Should not it be easier just to make Task.sleep instead of this hard logic?
what's the difference, and don't you think that managing a task and restarting the timer is more problematic than just using subject + debounce?

Adamant/Modules/ChatsList/ChatListViewController.swift
art-divin: The confusion comes from typealias "ObservableSender" is actually a PassthroughSubject. So... I would definitely never use typealias 😄

Adamant/Modules/ChatsList/ChatListViewController.swift
```diff
@@ -785,7 +799,15 @@ extension ChatListViewController: NSFetchedResultsControllerDelegate {
             tableView.endUpdates()
 
         case let c where c == unreadController:
```
art-divin, line 801: just for the sake of clarity, we can rename `c` to `controller` as intended

art-divin (approved):

PR #884, opened 2025-06-18
[trello.com/c/SL4zLZnb] Show "❌" when the transaction is failed instead of clock image

Lainaaa:


vovameister (commented):
why do you use static let and TransactionDetailsViewControllerBase.failedValueString instead of common let and self?.failedValueString?

Lainaaa:
> why do you use static let and TransactionDetailsViewControllerBase.failedValueString instead of common let and self?.failedValueString?

I just followed existed code style

vovameister (approved):

PR #891, opened 2025-07-26
[trello.com/c/pRsTtPYQ] marked each failed files

### LINKED ISSUE #921: [Bug, UX/UI] Incorrect display of unavailable attachments in chat
### Summary

When a chat includes messages with unavailable attachments, they are displayed incorrectly.

### Steps to reproduce

1. Open a chat containing messages with unavailable attachments (chat with ..55297 in test account)
2. Tap on the **red circle with arrow** icon near the message
3. Observe the incorrect display/behavior

### Expected behavior

- On tap, Icon behavior should follow this sequence:
  - **Red circle with arrow** → **Clock (downloading)** → **Red circle with arrow** (if download fails again)  
- Show a **temporary popup** (Popup should disappear automatically after a short delay) if downloading fails again
  - Two kinds of message depending whether all files unavailable or some files:
    - All files
      - `Files are not currently accessible`
      - `Файлы недоступны`
    - Some files
      - `Some files are not currently accessible`
      - `Некоторые файлы недоступны`
  - Style:
Image

- Add **vibration feedback** (“Error” type)

### Actual behavior

On "Unavailable icon" tap, Nothing happen

### Screenshots or videos

Image

### Environment

**Platform:** iOS app  
**Version:** Build 3.11.0 (20)

vovameister:
most of the time I was trying to figure out what was wrong with the files. It turned out that the message simply contained the ID of only the very first file and assigned it to all the others.
now to display this in the ui I made a check whether a specific id had already been used, this may look like a hard code, but the bug itself is very strange and our architecture with models of individual files does not really suggest that the service could check this when loading. it simply searches for files by 1 and the same id, although in fact all the files that were needed have already been downloaded.

### DIFF
--- a/Adamant/Modules/Chat/View/Subviews/ChatMedia/Content/Views/ChatFileContainerView/FileListContentView.swift
+++ b/Adamant/Modules/Chat/View/Subviews/ChatMedia/Content/Views/ChatFileContainerView/FileListContentView.swift
@@ -14,6 +14,7 @@ final class FileListContentView: UIView {
     private lazy var iconImageView: UIImageView = UIImageView()
     private lazy var downloadImageView = UIImageView(image: .asset(named: "downloadIcon"))
     private lazy var videoIconIV = UIImageView(image: .asset(named: "playVideoIcon"))
+    private var usedFileIds: Set<String> = []
 
     private lazy var spinner: UIActivityIndicatorView = {
         let view = UIActivityIndicatorView(style: .medium)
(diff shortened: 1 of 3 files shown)


Lainaaa:
Thanks for the explanation! 
Overall it should work, but what do you think to create an additional mini-service for such tasks? It would be much easier to understand in the feature where is the problem and maybe in the future it would be possible to include this service to the tests also

vovameister:
> Thanks for the explanation! Overall it should work, but what do you think to create an additional mini-service for such tasks? It would be much easier to understand in the feature where is the problem and maybe in the future it would be possible to include this service to the tests also

I'm not sure that makes sense
in my understanding, if we add a separation of responsibility, we need to refactor the architecture of working with messages containing files, for which we currently do not have the resources

Lainaaa (approved):

Issue #910, opened 2025-09-23
[Feat, Enhancement] Improve in-chat scroll position | "New messages" separator


dev-adamant-im:
### Summary

Implement consistent behavior for in-chat scroll position, the **Scroll to bottom** button, and the **New messages** separator across all entry points

### Motivation

Goal: predictable UX for reading and catching up on messages.

Users expect a chat app to restore their reading position accurately, show where new messages start, and provide a clear way to jump to either the newest or the first unread message.  
Current behavior may be inconsistent, especially when switching between chats or receiving new messages while inside a chat.

### Detailed description

Follow Specification: In-chat scroll position & New messages separator https://github.com/orgs/Adamant-im/discussions/24

## Acceptance Criteria

- Scroll state is saved/restored per chat consistently  
- “New messages” separator appears only in correct cases  
- Scroll to bottom button behaves as described  
- Edge cases covered (short chats, multiple notifications, entering from other screens)  

### Environment
- [ ] Platforms: iOS, macOS
- [ ] States: Logged in, multiple chats with varied history
- [ ] Test data: Chats with 0/short/long histories; unread counts; images/files/long text
- [ ] Network: Stable, Slow 3G, Offline/Airplane, VPN on/off
- [ ] Orientation: Portrait & Landscape

### A. Scroll State Saving
**Precondition:** User scrolls within a chat.
- [x] A1. At the very bottom → **Do not save** position (button hidden)
- [x] A2. Few pixels from bottom (button still hidden) → **Do not save**
- [x] A3. Not at bottom (button visible) → **Save** position
- [x] A4. Short chat (< screen height) → Always considered bottom → **Do not save**

**Persistence**
- [x] A5. Navigate away (to Chat list) and back → Position restored
- [x] A6. App background → foreground → Position preserved
- [ ] A7. App restart → Position preserved per chat
- [x] A8. Switch to another chat → Return → Position restored

### B. “Scroll to bottom” Button Visibility & Action
- [x] B1. Show when last single-line message is ~¾ hidden
- [x] B2. No new messages + above separator → Button scrolls to **very bottom**
- [x] B3. New messages present → Button scrolls to **“New messages” separator**
- [x] B4. Already below separator (separator visible or below) → Button scrolls to **very bottom**
- [x] B5. Button counter increments with each incoming message while not at bottom
- [x] B6. Counter resets after button tap or manual scroll past separator

### C. “New messages” Separator Behavior
[...]

PR #928, opened 2025-12-08
Save chat position after reloading the app

Lainaaa:
## Description

Save chat position after restartring the app. Positions will be deleted on a logout.

## Related issue

[910](https://github.com/Adamant-im/adamant-iOS/issues/910)

## Notes for reviewers

Initially tried to reuse CoreData chatroom, but was not able to implement this approach and decided to use UserDefaults instead.


metalisk (approved):

adamantmm:
@Lainaaa 

Do we want this feature?
I think no.

At least now, when we don't even store chats on device (another big task, but not for now).

Let's close the pr without merging.

Lainaaa:
> Do we want this feature? I think no.
> 
> At least now, when we don't even store chats on device (another big task, but not for now).
> 
> Let's close the pr without merging.

I also was not sure about it, lets close for now.

